PR #168, opened 2026-04-16
fix(dev): exec claude in wrapper for Warp terminal integration

ryanrozich:
## Summary
- Replace child-process `claude "$@"` with `exec claude "$@"` in `catalyst-claude.sh` so the process image becomes `claude` — restoring Warp's rich sidebar (repo, branch, changes) and notification integration
- Move heartbeat + cleanup into a background watcher that polls the wrapper PID and records session end when claude exits
- Tradeoff: session end status is always `done` (watcher can't capture exit code from exec'd process)

Closes CTL-82

## Test plan
- [ ] Launch `catalyst-claude` in Warp — verify sidebar shows Claude-specific info (repo path, branch, change count)
- [ ] Verify Warp notifications fire for Claude messages
- [ ] Check `catalyst-session list --active` shows heartbeats during session
- [ ] Exit claude and verify session is recorded as ended (`catalyst-session history --limit 1`)
- [ ] Verify `.catalyst/.session-id` is cleaned up after exit
- [ ] Test fallback: rename `catalyst-session.sh` temporarily — verify `catalyst-claude` still launches claude directly

### DIFF
--- a/plugins/dev/scripts/catalyst-claude.sh
+++ b/plugins/dev/scripts/catalyst-claude.sh
@@ -16,8 +16,10 @@
 #   1. Detects ticket from --ticket flag, .catalyst/workflow-context, or branch
 #   2. Starts a catalyst session (records PID, working dir, branch)
 #   3. Exports CATALYST_SESSION_ID so skills inside claude inherit it
-#   4. Runs a background heartbeat every 60s to keep updated_at fresh
-#   5. On exit, calls catalyst-session end with done/failed status
+#   4. Spawns a background watcher for heartbeat + cleanup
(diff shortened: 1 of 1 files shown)


ryanrozich:
Good catch on the missing `disown` — fixed in a follow-up.

The watcher works correctly today because `exec` replaces the process image (bash's job table vanishes) and the subshell already has `trap '' INT TERM`. But `disown` after `) &` is cheap defensive practice: it explicitly removes the job from bash's job table before `exec`, preventing any edge case where bash might send SIGHUP to background jobs, and it documents the intent that this process should outlive the shell.

Follow-up: #171

PR #435, opened 2026-05-05
fix(dev): correct webhookTunnel field and add smee-client dependency

ryanrozich:
## Summary

Two bugs that caused orchestrations to always fall back to REST polling instead of event-driven waits:

- Skills (`wait-for-github`, `oneshot`, `monitor-events`) read `.webhookTunnel.state` (non-existent field) instead of `.webhookTunnel.connected`. The jq fallback always returned `"unknown"`, which never equaled `"running"`, so `USE_REST=true` was set unconditionally on every invocation.
- `smee-client` was missing from `orch-monitor/package.json`, so the webhook tunnel silently failed to start on every monitor launch with `Cannot find package 'smee-client'`.

## Changes

- `catalyst-monitor.sh` — add `bun.lock -nt node_modules` staleness check so future package additions trigger reinstall even when `node_modules` already exists
- `orch-monitor/package.json` + `bun.lock` — add `smee-client ^5.0.0` to dependencies
- `skills/wait-for-github/SKILL.md` — new reference skill (correct `.webhookTunnel.connected` field)
- `skills/monitor-events/SKILL.md` — update to correct `.webhookTunnel.connected` field


ryanrozich:
### Code review

Found 1 issue:

1. **Incorrect filter for `check_suite.completed` in Quick Reference** — the PR replaces `.detail.prNumbers // [] | contains([N])` with `.scope.pr == ${PR_NUMBER}`, but the `check_suite` envelope is built with `scope: { repo: event.repo }` only — there is no `scope.pr` field on check_suite events. The PR number for check_suite events lives in `.detail.prNumbers[]`, so the old filter was correct and the new one will silently never match.

https://github.com/coalesce-labs/catalyst/blob/0d2a0feac37f7cdde678ef2c61d12e5955ac732a/plugins/dev/skills/wait-for-github/SKILL.md#L183

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1352, opened 2026-06-06
feat(dev): add catalyst.dispatch_mode telemetry dimension (OTL-7)

ryanrozich:
## What & why

Phase events from the **execution-core daemon** and from **/orchestrate** are currently indistinguishable — both emit `task.type=phase-<P>` + `catalyst.exec_context=phase-bg`. This adds a `catalyst.dispatch_mode` resource attribute so phase cost/token/tool metrics can be segmented by **who dispatched them**.

Emit side only — the consumer side (collector label `catalyst_dispatch_mode` + a `sum by(catalyst_dispatch_mode)` dashboard panel) ships separately in `coalesce-labs/catalyst-otel` (OTL-7).

## Design — pure explicit-var, no env inference

A multi-agent adversarial review (3 independent lenses: completeness, correctness, collision/naming) **rejected** the original "infer `execution-core` from `CATALYST_EXECUTION_CORE`" approach as fragile to env inheritance (an execution-core child that re-dispatches, or a future caller, could be silently mislabeled). The implemented design is therefore **purely explicit**:

| Site | Change |
|---|---|
| `execution-core/dispatch.mjs:73` | set `CATALYST_DISPATCH_MODE="execution-core"` in the spawned phase-agent env (sibling to the existing `CATALYST_EXECUTION_CORE:"1"`) |
| `orchestrate-dispatch-next:~436` | set `CATALYST_DISPATCH_MODE="orchestrate"` in the phase-agent-dispatch inline env prefix |
| `phase-agent-dispatch` `compose_otel_resource_attrs()` | read the var, validate against the closed enum `{execution-core\|orchestrate\|oneshot}`, append `catalyst.dispatch_mode` **only when valid** — else **omit** (honest unknown) |

Benefits: no inference ⇒ unknown callers are never mislabeled; the whitelist blocks a typo'd producer from injecting an arbitrary Loki/Prometheus label; and **every existing pinned OTEL-string test stays valid unchanged** (they set no `CATALYST_DISPATCH_MODE`, so nothing is appended).

## `oneshot` is reserved (decision for review)

`oneshot` is in the enum for forward-compat but **not actively emitted**: `/oneshot` runs phases inline in an interactive session, already distinguishable via `exec_context=interactive` + `task.type=oneshot`, so it never reaches the bg dispatch path. If you'd prefer to also stamp launcher sessions (a "cost by dispatch model across launchers" view), that's a clean follow-up — flagging for your call.

## Naming note (for review)

Telemetry attr `catalyst.dispatch_mode` / env `CATALYST_DISPATCH_MODE` (enum: execution-core\|orchestrate\|oneshot) is distinct from the pre-existing config `catalyst.orchestration.dispatchMode` (enum: phase-agents\|oneshot-legacy\|execution-core). Different casing/namespace; the attr name matches the OTL-7 handoff. Happy to rename (e.g. `launch_mode`) if you think the collision risk warrants it.

## Tests (all green locally — note: no CI job runs these suites)

- `phase-agent-dispatch.test.sh` (207 pass): execution-core/orchestrate tag the attr; absent var omits it **even with `CATALYST_EXECUTION_CORE=1`** (no inference); invalid value rejected by whitelist. `fresh_env` now unsets both vars for hermeticity.
- `dispatch.test.mjs` (34 pass): daemon spawn carries `CATALYST_DISPATCH_MODE=execution-core` alongside `CATALYST_EXECUTION_CORE=1`.
- `orchestrate-dispatch-next.test.sh` (140 pass): wave sets `dispatch_mode=orchestrate` and does **not** leak the daemon fencing token (non-leak contract).
- `shellcheck -S error` clean; `bash -n` / `node --check` clean; `audit-references --ci` exit 0.

Ref handoff `2026-06-06_07-14-58_catalyst-telemetry-dimensions-and-headless-emission.md`.

### DIFF
--- a/plugins/dev/scripts/__tests__/orchestrate-dispatch-next.test.sh
+++ b/plugins/dev/scripts/__tests__/orchestrate-dispatch-next.test.sh
@@ -464,6 +464,12 @@ phase_agent_dispatch_setup() {
 	cat >"${SCRATCH}/bin/phase-agent-dispatch" <<'EOF'
 #!/usr/bin/env bash
 echo "$@" >> "$PHASE_DISPATCH_LOG"
+# OTL-7: capture the dispatch-mode env so tests can assert the producer (the
+# /orchestrate wave) tagged this invocation and did NOT leak the daemon token.
+{
+  echo "CATALYST_DISPATCH_MODE=${CATALYST_DISPATCH_MODE:-}"
+  echo "CATALYST_EXECUTION_CORE=${CATALYST_EXECUTION_CORE:-}"
+} >> "${PHASE_DISPATCH_ENV_LOG:-/dev/null}"
 # Write the per-phase signal so the dispatcher's idempotency check sees it on
 # subsequent invocations — mirrors what the real helper does.
 ORCH_DIR=""; PHASE=""; TICKET=""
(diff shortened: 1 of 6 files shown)


ryanrozich:
Closing as superseded — catalyst.dispatch_mode telemetry already live on main via CTL-1008 (#1793, merged 2026-06-11). Note: this 282-behind branch had a stricter whitelist + explicit producer-side values not on main; if that hardening is wanted, file a small follow-up against current main rather than reviving this branch.

PR #1850, opened 2026-06-11
fix(dev): CTL-1050 service health — broker/exec-core show green when daemons are live

ryanrozich:
## Summary

- Fix `RECENCY_MATCHERS` to use `catalyst.broker`/`catalyst.execution-core` (the actual service names the catalyst event log uses, prefixed with `catalyst.`) instead of bare `broker`/`execution-core` — fixes false-down on both daemons
- Fix OTel collector recency fallback to infer liveness from Loki status (Loki up → collector up) instead of searching for `claude-code` events that never appear in the local catalyst event log
- Add regression tests for both fixes (8 total, all green)

## Test plan

- [x] `bun test __tests__/service-health-monitor.test.ts` — 8 pass, 0 fail
- [x] Live: `/api/health/services` on mini shows broker/execution-core as green after merge+deploy

### DIFF
--- a/plugins/dev/scripts/orch-monitor/__tests__/service-health-monitor.test.ts
+++ b/plugins/dev/scripts/orch-monitor/__tests__/service-health-monitor.test.ts
@@ -57,6 +57,48 @@ describe("readEmissionAge", () => {
   });
 });
 
+describe("broker / execution-core recency uses catalyst.* service names", () => {
+  it("broker shows up when catalyst.broker emitted recently", async () => {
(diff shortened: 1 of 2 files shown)


ryanrozich:
Closing as superseded — CTL-1050 service-health fix already landed via #1912 (CTL-1087, merged 2026-06-12, commit 2b1eae58). Post-rebase lib/service-health-monitor.ts is byte-identical to main and main's tests are a superset. No net change.

PR #1991, opened 2026-06-14
feat(dev): CTL-1127 — inbox row: restore favicon, unwrap ID, two-line title, drop verb cluster, fix pane accent

ryanrozich:
08660924 feat(dev): CTL-1127 — inbox row: restore favicon, unwrap ID, two-line title, drop verb cluster, fix pane accent

Refs: CTL-1127


ryanrozich:
Rebased onto current main (was 305 behind); the only conflict was a one-line import union in `entity-icon.test.ts`. @codex review

PR #2245, opened 2026-06-18
fix(dev): CTL-1273 — one roster source of truth (no reader/writer divergence)

ryanrozich:
## CTL-1273 — one roster source of truth

Collapses the cluster roster reader/writer divergence so a stale or absent secondary copy can never silently drop a node or shadow a `cluster add`.

### Problem
`getClusterHosts()` (`config.mjs`) read the control-plane `cluster.json.roster` **first** and **else** the project `hosts.json`, while every writer (`cli/cluster.mjs` add/remove via `getCatalystRepoDirHostsPath`) only ever wrote the project `hosts.json`. Nothing in the codebase writes `cluster.json.roster` — it arrives only via `cluster-sync` git pull. So the two were structurally divergent: once a cluster clone existed, a `cluster add` was silently shadowed, and a stale `cluster.json` (`["mini"]`) silently dropped `mini-2` fleet-wide.

### Fix (degrade-safe option b from the design doc)
Reconcile the two enrollment sources with a **UNION** instead of cluster-first-replace: the active roster = enrolled in `cluster.json` OR the project `hosts.json`. The CLI write target stays the project `hosts.json` on purpose, so it is now **always** honored by the reader and no source can silently subtract a node. Divergence between the two is made **visible** via a single non-load-bearing `log.warn` (gated to both-non-empty-and-differ, so identical sources never warn). The `schemaVersion` too-new degrade is preserved (a too-new `cluster.json` contributes nothing to the union).

### Live-fleet safety
No-op on the running fleet: the cluster clone is absent on both hosts, so `clusterRoster=[]` and the active roster stays the project `hosts.json` = `["mini","mini-2"]`, byte-identical to today.

### Trade-off (documented in the function comment)
The union is additive — a node removed from `cluster.json` but lingering in a stale project `hosts.json` is re-added. This is deliberate: under the enrollment-vs-liveness redesign, "node offline" is liveness (CTL-L), not a roster edit, so over-inclusion is harmless. Authoritative subtraction is a separate decision (CTL-1214).

### Tests
- `config.test.mjs`: new `getClusterHosts roster reconcile (CTL-1273)` suite (stale-drop, project-only-visible, union order+dedup, too-new contributes nothing, empty-everything default, divergence warn, identical-no-warn). The existing CTL-1211 cluster-repo-first suite stays green unchanged (backward-compatible).
- `cli-cluster.test.mjs`: new `reader/writer agree end-to-end (CTL-1273)` suite proving a `runAdd` is immediately visible to `getClusterHosts` even with a stale `cluster.json` present.
- `bun test config.test.mjs cli-cluster.test.mjs` → 141 pass / 0 fail. `doctor.test.mjs` → 46/0 green.

Notes: autonomous wave-1 impl, **NEEDS REVIEW before merge**. References CTL-1273.

### DIFF
--- a/plugins/dev/scripts/execution-core/cli-cluster.test.mjs
+++ b/plugins/dev/scripts/execution-core/cli-cluster.test.mjs
@@ -1,7 +1,7 @@
 // cli-cluster.test.mjs — CTL-1188. Unit tests for cli/cluster.mjs.
 // Run: cd plugins/dev/scripts/execution-core && bun test cli-cluster.test.mjs
 import { describe, test, expect, beforeEach, afterEach } from "bun:test";
-import { mkdtempSync, writeFileSync, readFileSync, rmSync, existsSync } from "node:fs";
+import { mkdtempSync, writeFileSync, readFileSync, rmSync, existsSync, mkdirSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
@@ -11,7 +11,10 @@ import {
   renameHost,
   setAnchor,
   tune,
+  runAdd,
+  runRemove,
 } from "./cli/cluster.mjs";
+import { getClusterHosts } from "./config.mjs";
 
 // ── Helpers ────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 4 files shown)


ryanrozich:
## Adversarial review — CTL-1273 (one roster source of truth)

**Verdict: PASS** — correct, in-scope, well-tested, no reward-hacking, no existing test weakened.

### Correctness (vs. design doc concern A + both Gherkin scenarios)
- `getClusterHosts()` reworked from cluster-FIRST-replace → **UNION** of `cluster.json.roster ∪ project hosts.json`, order-stable (cluster first), de-duplicated. This satisfies both acceptance scenarios:
  - *"Adding a node takes effect"* — the CLI writer target (`getCatalystRepoDirHostsPath` → project `hosts.json`) is now always honored by the reader; it can no longer be shadowed by a stale/absent `cluster.json`.
  - *"A stale roster copy cannot silently override the live one"* — union is additive, so no enrollment source can silently SUBTRACT a node. Verified: stale `cluster.json=["mini"]` + project `["mini","mini-2"]` → `["mini","mini-2"]` (pre-fix returned `["mini"]`, the live landmine).
- Schema policy preserved: too-new `cluster.json` contributes nothing to the union (degrade to project roster).
- **Live-fleet safe**: cluster clone absent here → `clusterRoster=[]` → union == project roster `["mini","mini-2"]`, byte-identical to prior behavior (verified `.catalyst/hosts.json` + absent `~/catalyst/catalyst-cluster`).

### Scope — clean
- Only the 4 declared files touched. `getCatalystRepoDir()` (CTL-1271 / concern B's cwd surface) is **called but not modified**. No scheduler/dispatch changes. Did not overreach into CTL-1214's "extract roster → fleet config" surface — it reconciled the two existing sources, no new fleet-config file.

### Tests — adequate, non-vacuous, RED-first verified
- Independently confirmed RED-first by running the new tests against `origin/main`'s `config.mjs`: **4 config cases fail** (stale-drop, project-only-visible, union-order, divergence-warn) + **1 cli-cluster e2e fails** (stale-`cluster.json` shadows the add). All green on branch.
- Assertions are concrete (`toEqual`/`toContain`/`not.toContain`), order+dedup proven by `["mini","mini-2","mac-studio"]`. The warn spy is a real injected `log.warn` override restored in `finally`; divergent → `seen.length >= 1`, identical → `seen.length === 0` (gating proven both ways).
- Full suites pass: config.test + cli-cluster.test = 141 pass / 0 fail; doctor.test guard 46 pass / 0 fail.
- No existing test weakened. Note: the CTL-1211 case *"cluster.json roster wins over project hosts.json"* still passes, but its assertion is now satisfied by the union coincidentally (its name is mildly stale) — assertion intact, no regression.

### Reward-hacking — none
- `.mjs` files (no `as any`/`@ts-ignore`). No non-null `!`, no `.only`/`skip`, no `toBeTruthy()`/`expect(true)`, no empty assertions.

### Minor finding (low)
- The divergence `log.warn` fires **per scheduler tick** (~30s) whenever both sources are non-empty and disagree (`getClusterHosts` is called per-tick at `scheduler.mjs:2773` / `monitor.mjs:573`). During a sustained misconfig (e.g. an un-pulled enrollment copy) this is repeated stderr output rather than a once-per-process warning. Acceptable given the design's "make divergence visible, not silent" intent, stderr-only + non-load-bearing, and the cluster-active path being dormant fleet-wide today — but worth a follow-up de-dup latch if/when the cluster repo path is activated.

🤖 Adversarial review by Claude Opus 4.8

ryanrozich:
Superseded. The fleet roster is moving off per-repo files entirely onto the **Linear cluster anchor** (CTL-1273), and **Tailscale is dropped** from the design (open-source: it must not be required just to list cluster members). The single-host default needs nothing; Linear is already a hard dep for multi-host. This branch's *union of per-repo copies* approach is obsolete — there are no per-repo copies anymore. Reworked under the reshaped CTL-1273. Branch kept for reference.

PR #2246, opened 2026-06-18
fix(dev): CTL-1271 — deterministic roster path + loud boot assertion

ryanrozich:
## CTL-1271 — deterministic roster path + loud boot assertion

Hardens the execution-core daemon so it can **never silently shrink to a one-node cluster** when it can't find its real roster (this is how mini-2 was knocked out of the fleet). Conservative: changes WHERE the roster file is found, not the roster-precedence logic.

### Three coordinated changes
1. **`config.mjs` `getCatalystRepoDir()`** stops trusting bare `process.cwd()`. Resolves `<repoRoot>/.catalyst` from a deterministic anchor: `CATALYST_CONFIG_FILE` → `CATALYST_REPO_ROOT` → (CLI-writer only) cwd → `null` sentinel. `getClusterHosts()` treats `null` as "no project roster" → single-host default, so a mis-launched daemon never silently reads a decoy `hosts.json`. `getCatalystRepoDirHostsPath()` (the CLI write target, always run in-repo) opts into the cwd anchor to preserve its behavior unchanged.
2. **`daemon.mjs` `main()`** resolves `configPath` via a new exported pure helper `resolveDaemonConfigPath({env, listProjects, cwd})` (env → first registry repoRoot → flagged cwd last resort), never bare `process.cwd()`. The boot path adds `{roster, sourcePath, multiHost}` to the `daemon started` log and emits a single **LOUD `log.warn`** (fail-open, never throws mid-boot) when resolution was non-deterministic (cwd) OR the roster degraded to single-host while multi-host was expected. A legitimately single-host deployment stays silent.
3. **`catalyst-execution-core` `cmd_start`** pins a canonical `CATALYST_CONFIG_FILE` / `CATALYST_REPO_ROOT` (derived from the registry's first `repoRoot`) before `nohup`, so cwd can never decide the roster. Best-effort: a failed derivation warns but never blocks launch.

### Scope discipline
Strictly avoids `getClusterHosts()` precedence (**CTL-1273**) and `hostMembershipWarning` internals (**CTL-1272**) — only reads them.

### Tests
- `config.test.mjs`: new `getCatalystRepoDir deterministic anchor (CTL-1271)` describe — cwd-decoy never read, `CATALYST_REPO_ROOT` anchor wins over cwd, unresolved → `null` sentinel (not the cwd roster), CLI-writer cwd target preserved.
- `daemon.test.mjs`: extends the CTL-862 blocks — `resolveDaemonConfigPath` (registry/env/cwd) cases; boot log carries `sourcePath`+`multiHost`; single-host-with-multi-host-expectation **warns loudly**; cwd resolution **warns loudly**; legitimate single-host does **NOT** warn. All pre-existing CTL-862/CTL-1211 tests preserved.

**Test command:** `cd plugins/dev/scripts/execution-core && bun test config.test.mjs daemon.test.mjs` → **220 pass, 0 fail**. Dependent suites (`cli-cluster`, `join-bundle`, `doctor`, `cluster-ownership`, `cluster-heartbeat-publisher`, `recovery-pass-context`) → **113 pass, 0 fail**.

Design: `thoughts/shared/plans/2026-06-18-cluster-membership-liveness-redesign.md` (concern B). References CTL-1271.

> Note: autonomous wave-1 implementation. **NEEDS REVIEW before merge.**

### DIFF
--- a/plugins/dev/scripts/catalyst-execution-core
+++ b/plugins/dev/scripts/catalyst-execution-core
@@ -166,6 +166,28 @@ cmd_start() {
 		fi
 		unset _ocid _ocsec _otok
 	fi
+	# CTL-1271: pin a DETERMINISTIC roster anchor before nohup so the daemon's
+	# cwd can NEVER decide the cluster roster. The daemon resolves its roster from
+	# <repoRoot>/.catalyst/hosts.json; if it is launched from the wrong directory
+	# with no CATALYST_CONFIG_FILE, it silently falls back to a one-node cluster
+	# and disables HRW work-sharing with no error (this is how mini-2 was knocked
+	# out of the fleet). Derive the canonical config path from the registry's first
+	# enrolled project repoRoot and export it (plus CATALYST_REPO_ROOT as a
+	# secondary anchor). Best-effort: a failed derivation must NEVER block launch —
+	# the daemon's resolveDaemonConfigPath repeats this derivation and warns loudly
+	# if it cannot resolve deterministically. Honor any value already set.
+	if [[ -z "${CATALYST_CONFIG_FILE-}" ]]; then
+		local _registry="$EC_DIR/registry.json" _repo_root
+		_repo_root="$(jq -r '.projects[0].repoRoot // empty' "$_registry" 2>/dev/null)"
+		if [[ -n "$_repo_root" && -d "$_repo_root/.catalyst" ]]; then
+			export CATALYST_CONFIG_FILE="$_repo_root/.catalyst/config.json"
+			export CATALYST_REPO_ROOT="$_repo_root"
+			echo "catalyst-execution-core: pinned roster anchor CATALYST_CONFIG_FILE=$CATALYST_CONFIG_FILE (CTL-1271)" >&2
+		else
+			echo "catalyst-execution-core: WARNING could not derive a canonical CATALYST_CONFIG_FILE from $_registry — daemon will assert its roster source at boot (CTL-1271)" >&2
+		fi
+		unset _registry _repo_root
+	fi
 	nohup "$runtime" "$DAEMON_SCRIPT" --pid-file "$PID_FILE" >"$LOG_FILE" 2>&1 &
 	local server_pid=$!
 	disown "$server_pid" 2>/dev/null || true
--- a/plugins/dev/scripts/execution-core/config.mjs
+++ b/plugins/dev/scripts/execution-core/config.mjs
@@ -157,17 +157,37 @@ export function getLayer2ConfigPath() {
   );
 }
 
-// The repo root that owns the committed cluster roster (.catalyst/hosts.json).
-// CATALYST_CONFIG_FILE points at <repoRoot>/.catalyst/config.json (mirrors the
-// reaper-config resolution in daemon.mjs main()); otherwise fall back to the
-// daemon's cwd. Re-resolved per call so tests can redirect via the env var.
-function getCatalystRepoDir() {
+// getCatalystRepoDir — the <repoRoot>/.catalyst directory that owns the committed
+// cluster roster (hosts.json). CTL-1271: this must resolve from a DETERMINISTIC
+// anchor, never bare process.cwd(), or a daemon launched from the wrong directory
+// silently reads a wrong/absent hosts.json and shrinks to a one-node cluster with
+// no error and no log (exactly how mini-2 was knocked out of the fleet).
+//
+// Precedence (re-resolved per call so tests redirect via the env vars):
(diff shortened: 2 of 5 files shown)


ryanrozich:
## Adversarial review — CTL-1271 (deterministic roster path + loud boot assertion)

**Verdict: PASS** ✅

Reviewed the diff against the design doc (concern B) and both ticket acceptance scenarios. The change is conservative, correctly scoped, and well-tested.

### Correctness
- **Root-cause fix is sound.** `getCatalystRepoDir()` now resolves from a deterministic anchor (`CATALYST_CONFIG_FILE` → `CATALYST_REPO_ROOT` → null sentinel) instead of bare `process.cwd()`. The reader (`getClusterHosts`) treats `null` as "no project roster" → single-host default, and never reads a cwd-relative decoy `hosts.json`. The CLI write target (`getCatalystRepoDirHostsPath`) opts into cwd via `allowCwd:true`, preserving the CTL-1211 reader/writer divergence guard.
- **Backward-compat verified.** When `CATALYST_CONFIG_FILE` is set (normal production), `getCatalystRepoDir()` returns a non-null dir → the `if (repoDir)` guard reads `hosts.json` exactly as before. Identical behavior.
- **`resolveDaemonConfigPath`** is a clean pure helper (env → registry repoRoot → flagged cwd), unit-tested with injected env/registry/cwd. main() now uses it; the cwd path is the only one flagged `"cwd"` so the boot path can warn.
- **Boot assertion fires after the CTL-862 back-write** (line 475), so the production `sourcePath` computation reads the resolved `CATALYST_CONFIG_FILE`. Scenario 2 (boot log states roster + sourcePath + multiHost) is satisfied.
- **Launcher** pins `CATALYST_CONFIG_FILE`/`CATALYST_REPO_ROOT` from the registry's first repoRoot before nohup, best-effort, never blocks launch — matches the existing in-function pattern (`jq ... // empty 2>/dev/null`, warn-on-failure, `unset` at end). jq path `.projects[0].repoRoot` matches the registry shape and `EC_DIR` is correct.

### Scope
Strictly in-bounds. `getClusterHosts()` precedence (cluster-repo-first / CTL-1273) is untouched; `hostMembershipWarning` internals (CTL-1272) untouched; scheduler/hrw untouched. Exporting the previously-private `getCatalystRepoDir` is additive and safe (confirmed no other importer relied on its absence).

### Test adequacy (verified by mutation, not just by reading)
- Reverting the boot-assertion conditions to `false` kills exactly the two loud-warn tests (single-host-with-multihost-expectation, cwd-non-deterministic) — genuine regression guards.
- Reverting `getCatalystRepoDir` to the bare cwd fallback kills the two deterministic-anchor config tests (decoy roster leaks in) — genuine guards.
- The "legitimately single-host stays silent" test correctly asserts no warn with deterministic resolution + no multi-host expectation (mirrors `hostMembershipWarning`'s length≤1 → null discipline, so the assertion never becomes noise).
- No existing test lines removed (test diff is purely additive). Pre-existing CTL-862 boot-log test preserved unchanged.
- All suites green: config.test + daemon.test 220 pass / 0 fail; dependent cli-cluster + cluster-heartbeat-publisher + cluster-claim + doctor + recovery-pass-context 121 pass / 0 fail; join-bundle 12 pass / 0 fail.

### No reward-hacking
No `as any` / `@ts-ignore` / non-null `!` / void tricks / empty assertions in the diff.

### Minor notes (non-blocking, no action required)
1. **`expectedMultiHost` is never wired into production `main()`** — so in production the only loud-warn path is the cwd-fallback case. This is acceptable: the deterministic resolution itself fixes the bug at the root (launcher/registry resolve to "env"/"registry", reading the real roster), and the `degradedSingleHost` branch is a tested seam available to callers. A *stale-content* roster (committed `["mini"]`) would still single-host silently, but that is concern A / CTL-1273, explicitly out of scope here.
2. **Launcher jq uses `.projects[0]` (raw first entry)** while `resolveDaemonConfigPath` uses `projects.find(p => p.repoRoot)` (first valid entry). In a malformed-registry edge case (entry 0 missing repoRoot) the launcher would skip-and-warn while the daemon resolves deterministically to entry 1 — a harmless divergence; both are deterministic and the launcher explicitly defers to the daemon's boot-time resolution.

🤖 Adversarial review by Claude

ryanrozich:
Superseded. The fleet roster is moving off per-repo files entirely onto the **Linear cluster anchor** (CTL-1273), and **Tailscale is dropped** from the design (open-source: it must not be required just to list cluster members). The single-host default needs nothing; Linear is already a hard dep for multi-host. This branch's *deterministic file-path resolution* is moot — there is no roster file to resolve. The boot-assertion value survives in the reshaped CTL-1271 (assert the anchor-resolved roster). Branch kept for reference.

PR #2247, opened 2026-06-18
feat(dev): CTL-1272 — Tailscale + peer-HTTP liveness source (additive)

ryanrozich:
## CTL-1272 — Tailscale + peer-HTTP liveness source (additive)

Builds a NEW, **additive** liveness source — Tailscale device presence + a peer-HTTP `/healthz` check over the tailnet — exported as a clean seam (`node-presence.mjs`) that **CTL-1091's** `liveness-roster.mjs` will later consume (swap its input from `readPeerHeartbeats` to this presence probe).

This ticket is dormant: it does **NOT** touch the dispatch path, HRW, the soft-CAS claim, or the existing Linear-attachment heartbeat (`cluster-heartbeat*.mjs` left intact so nothing regresses — it may stay for dashboard display). `node-presence.mjs` has **no caller** in this PR; only the new read-only `/healthz` route is live.

### Surfaces
- **`plugins/dev/scripts/execution-core/node-presence.mjs` (NEW)** — pure + injectable (exec + fetch seams, mirroring `cluster-heartbeat`/`cluster-sync`'s "default real runner, override in opts" convention). Exports `readTailscaleStatus`, `tailscaleNameForRoster`, `tailscaleOnline`, `probePeerHealth` (three-tier reachable/up/healthy), `isHostLive`, `readLivePeers`.
  - **Name-mapping**: `nameMap` > DNSName-first-label (`mini.tail…`→`mini`) > `HostName` verbatim. Tags are **null fleet-wide** (verified live 2026-06-18: `RyansMini250233`/DNSName `mini`), so tag-based mapping is unusable this ticket — documented as a future enhancement.
  - **FAIL-OPEN** (the highest-risk boundary): a probe-infra error (exec spawn failure / null status / `/healthz` fetch throw while Online=true) degrades to **assume-live**, never mass-evicts. A **definitive** negative (host present with `Online===false`, or a clean `/healthz` reporting unhealthy/stale) is honored as not-live.
  - **Self** is always live and never probed (excluded via `getHostName()` — the laptop's TS HostName doesn't match a roster name).
  - **Hysteresis** (N-consecutive flip) is OUT of scope — it lives downstream in CTL-1091.
- **`plugins/dev/scripts/orch-monitor/server.ts`** — additive read-only `GET /healthz` returning `{ host, daemonAlive, lastTickAgeMs }`, reusing the existing `loadDaemonDeps`/heartbeat-memo/`deriveDaemonHealth` chain. Wired via a new optional injectable `healthReader` on `CreateServerOptions` (mirrors `daemonHealthReader`/`clusterReader`), mounted next to `/api/version`. No board read, no Linear/GitHub call, fail-open (never 5xx). `lastTickAgeMs` is heartbeat-derived (the daemon's liveness proxy), not a literal scheduler-tick counter.

### Tests
- `node-presence.test.mjs` — 24 hermetic cases (fixture modeled on the real fleet shape). `cd plugins/dev/scripts/execution-core && bun test node-presence.test.mjs` → 24 pass.
- `__tests__/healthz-endpoint.test.ts` — 4 cases (injected `healthReader` + production default). `cd plugins/dev/scripts/orch-monitor && bun test __tests__/healthz-endpoint.test.ts` → 4 pass.
- `server.ts` lint clean; parent typecheck shows zero new errors in `server.ts` (the only typecheck errors are pre-existing `ui/` module-resolution noise from uninstalled ui deps in the worktree).

> Autonomous wave-1 implementation — **NEEDS REVIEW before merge.** Linear ticket CTL-1272 left in Backlog. References CTL-1272; blocks CTL-1091.

### DIFF
--- a/plugins/dev/scripts/execution-core/node-presence.mjs
+++ b/plugins/dev/scripts/execution-core/node-presence.mjs
@@ -1,0 +1,302 @@
+#!/usr/bin/env node
+// node-presence.mjs — CTL-1272. NEW, ADDITIVE liveness source.
+//
+// Derives "is peer X up and healthy right now" from two real, already-running
+// fleet substrates — Tailscale device presence + a peer-HTTP /healthz check over
+// the tailnet — instead of the slow, rate-limited Linear-attachment heartbeat
+// (CTL-1090). Three tiers:
+//   1. reachable — Tailscale Online === true (network path exists)
+//   2. up        — the peer answers /healthz over the tailnet
+//   3. healthy   — that response reports a recent scheduler tick (not wedged)
+//
+// DORMANT in this ticket: this module has NO caller in the dispatch path. It is
+// the clean seam CTL-1091's liveness-roster.mjs (effectiveLiveRoster) will later
+// consume — swapping its input from readPeerHeartbeats to readLivePeers. The
+// existing Linear-attachment heartbeat (cluster-heartbeat*.mjs) is left intact so
+// nothing regresses; it may still feed the dashboard.
+//
+// HYSTERESIS (N-consecutive-observation flip) is OUT OF SCOPE here — it lives
+// downstream in CTL-1091's liveness-roster.mjs. This module exposes a single
+// point-in-time observation; the consumer wraps it with hysteresis.
+//
+// PURITY / INJECTABILITY (mirrors cluster-heartbeat.mjs / cluster-sync.mjs's
+// "default real runner, override in opts" convention): the Tailscale exec and
+// the HTTP fetch are both injectable so tests are hermetic — no real binary, no
+// real network. The ONLY config touch is getHostName() for self-exclusion in
+// readLivePeers (already env-redirectable via CATALYST_HOST_NAME). NO import of
+// scheduler.mjs / daemon.mjs.
+//
+// FAIL-OPEN (the highest-risk correctness boundary): a *probe-infrastructure*
+// error (tailscale binary missing / spawn failure / null status) must degrade to
+// "assume live" — never mass-evict the fleet over our own inability to see. A
+// *definitive negative*, however, is honored: a host present in Tailscale status
+// with Online===false, or a clean /healthz response reporting unhealthy, is a
+// real "not live".
+
+import { execFileSync } from "node:child_process";
+import { existsSync } from "node:fs";
+import { getHostName } from "./config.mjs";
+
+// macOS ships the binary inside the app bundle; PATH `tailscale` is the fallback
+// (Linux/CLI installs). Both are overridable via the exec seam.
+const TAILSCALE_APP_BUNDLE_BIN =
+  "/Applications/Tailscale.app/Contents/MacOS/Tailscale";
+
+// The monitor HTTP port the /healthz probe hits over the tailnet.
+const DEFAULT_MONITOR_PORT = 7400;
+
+// "Recent scheduler tick" staleness bound. The /healthz lastTickAgeMs is derived
+// from node.heartbeat freshness (~30s cadence), so a generous default tolerates
+// normal heartbeat jitter while still catching a wedged daemon. Overridable.
+const DEFAULT_STALE_MS = 120_000;
+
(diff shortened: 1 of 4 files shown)


ryanrozich:
## Adversarial review — CTL-1272 (Tailscale + peer-HTTP liveness source, additive)

**Verdict: PASS** (conservative, strictly-additive, dormant new liveness source)

### Scope — clean
- 4 files, all additive: `node-presence.mjs` (NEW), `node-presence.test.mjs` (NEW), `healthz-endpoint.test.ts` (NEW), `server.ts` (MODIFIED, **zero deletions** — verified `git diff` has no `-` lines, +65 only).
- Only import from `config.mjs` is `getHostName` (self-exclusion). Does **not** touch `getClusterHosts`/`getCatalystRepoDir` (CTL-1271/1273 surfaces), `scheduler.mjs` dispatch, HRW, the soft-CAS claim, or `cluster-heartbeat*.mjs`. No wave-1 sibling-branch collision.
- `node-presence.mjs` has **no caller in the dispatch path** — correctly dormant; it's the seam CTL-1091's `effectiveLiveRoster` will consume. Nothing regresses.

### Correctness
- Three-tier liveness (reachable / up / healthy) implemented per design §"Three-tier liveness".
- FAIL-OPEN is precise: probe-infra error (exec spawn fail / null status / fetch throw while Online=true) → assume-live; a definitive `Online===false` or a clean `/healthz` reporting unhealthy/stale → honored as not-live. This is the correct posture (never mass-evict on our own blindness).
- `productionHealth` in server.ts reuses the **same** `loadDaemonDeps()/heartbeatMemo/deriveDaemonHealth` chain as the existing `productionDaemonHealth` (verified: `heartbeatMemo.get(deps,{})` at server.ts:1415, mirroring :1382). `/healthz` is a 200 read-only report even when `daemonAlive:false` (never 5xx) — internally consistent with the probe contract.
- Minor note (non-blocking): a peer answering **HTTP 503** is classified `reachable:false` → `isHostLive` fail-opens to live. Since the shipped `/healthz` always returns 200 and signals unhealth in the JSON body, a 503 here only comes from infra, where fail-open is the correct conservative choice. Worth a future comment but not a defect.

### Tests — adequate, non-vacuous
- 28 tests pass (24 node-presence + 4 healthz). Each acceptance behavior is asserted: Online=false→not-live, reachable-but-wedged (stale tick)→not-live, both fail-open paths→live with `source:"fail-open"`, name-mapping end-to-end, and self-exclusion (`source:"self"` AND a `selfProbed===false` guard that genuinely proves self was skipped).
- Scenario 4 (hysteresis / no-flap) is explicitly OUT of scope, documented as living downstream in CTL-1091 — a legitimate scope boundary; this module is a single point-in-time observation.

### Reward-hacking — clean
- No `as any` / `as unknown as` / `@ts-ignore` / non-null `!`. The one `eslint-disable no-unused-vars` on `probePeerHealth(host,…)` is a documented intentionally-unused param for signature symmetry, not a type-error silencer. `void server?.stop(true)` in afterAll is standard async-cleanup discard. No weakened/deleted existing test.

### Test-suite note (non-regression)
- Full `bun test __tests__/` shows 11 fails / 4 errors, but every one is environmental in the isolated worktree (uninstalled ui/ deps: vite, @tanstack/react-router, marked; build:ui exit 127; a sqlite "file is not a database" fixture). The one suspicious case — `nav-signal-endpoints` "empty fleet / no anomaly" — **reproduces identically on pristine main (f6f847c3)**, so it is pre-existing, not caused by this diff.

ryanrozich:
Superseded. The fleet roster is moving off per-repo files entirely onto the **Linear cluster anchor** (CTL-1273), and **Tailscale is dropped** from the design (open-source: it must not be required just to list cluster members). The single-host default needs nothing; Linear is already a hard dep for multi-host. This branch's *Tailscale presence* source is dropped. The /healthz endpoint + hermetic peer-HTTP scaffolding here are salvageable for the reframed CTL-1272 (peer-to-peer health, no Tailscale), which is now parked as a follow-up. Branch kept for salvage.

PR #2249, opened 2026-06-18
CTL-1273: The fleet roster lives on the Linear cluster anchor (single source)

ryanrozich:
## What

Keystone of catalyst's cluster-membership redesign. Two coupled tickets, one branch/PR:

- **CTL-1273** — the fleet roster lives on the Linear **cluster anchor** issue (single source of truth: persistent `catalyst://node/<name>` attachments carrying `{name, address}`), never in a per-repo file.
- **CTL-1271** — the daemon announces its resolved roster + source at boot and never silently runs as a one-node cluster.

## Design

Roster resolver precedence in `getClusterHosts()`:
1. **Linear cluster anchor** — node-registration attachments when an anchor is configured (`getLivenessAnchorIssue()`).
2. **`static`** explicit node list from machine-local Layer-2 config (`catalyst.cluster.staticRoster`).
3. **Legacy `.catalyst/hosts.json`** — MIGRATION FALLBACK only (lowest priority).
4. **Single-host default** `[getHostName()]` — no Linear call when no anchor/static/hosts.json.

New module `node-roster.mjs` (enrollment/persistent) parallels `cluster-heartbeat.mjs` (liveness/transient). Injectable Linear client for hermetic tests; **FAIL-OPEN** (a Linear read error falls back to the next source, never silently empties the roster → mass-eviction).

CLI writer (`cli/cluster.mjs add/remove`) writes/deletes the node attachment on the **same anchor** the reader reads (structurally kills reader/writer divergence).

CTL-1271 boot assertion logs `{ roster, source, multiHost }`; loud operator-visible warn (not hard-refuse) when multi-host was expected but resolution yielded single-host; silent when single-host is the intent.

## Scope notes

- **Additive** — `.catalyst/hosts.json` is KEPT as a migration fallback. Deletion of the files + the CI guard that prevents reintroduction are an explicit follow-up, NOT this PR.
- Does not touch scheduler dispatch (CTL-1091) or liveness/cluster-heartbeat (CTL-1272, parked).

🤖 autonomous build — **NEEDS REVIEW**

### DIFF
--- a/plugins/dev/scripts/execution-core/cli-cluster.test.mjs
+++ b/plugins/dev/scripts/execution-core/cli-cluster.test.mjs
@@ -85,13 +85,14 @@ describe("buildStatus (CTL-1188)", () => {
 
 // ── addHost ────────────────────────────────────────────────────────────────
 
-describe("addHost", () => {
+describe("addHost (legacy hosts-fallback path, anchor:null)", () => {
   test("appends a new name and commits", () => {
     withRoster(["mini"], (hostsPath, read) => {
       const committed = [];
       const r = addHost("mac-studio", {
         hostsPath,
         self: "mini",
+        anchor: null,
         readPeers: () => ({}),
         git: (args) => committed.push(args),
         commit: true,
@@ -104,7 +105,7 @@ describe("addHost", () => {
 
   test("idempotent: name already in roster → code 2, no write", () => {
     withRoster(["mini"], (hostsPath, read) => {
-      const r = addHost("mini", { hostsPath, self: "mini", readPeers: () => ({}) });
+      const r = addHost("mini", { hostsPath, self: "mini", anchor: null, readPeers: () => ({}) });
       expect(r.code).toBe(2);
       expect(read()).toEqual(["mini"]);
     });
@@ -115,6 +116,7 @@ describe("addHost", () => {
       const r = addHost("ghost", {
         hostsPath,
         self: "mini",
+        anchor: null,
         readPeers: () => ({ ghost: { host: "ghost", last_seen: "2026-06-16T12:00:00Z" } }),
       });
       expect(r.code).toBe(2);
@@ -128,6 +130,7 @@ describe("addHost", () => {
       const r = addHost("mac-studio", {
         hostsPath,
         self: "mini",
+        anchor: null,
         readPeers: () => ({}),
         git,
         commit: false,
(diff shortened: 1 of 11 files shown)


ryanrozich:
## Adversarial review — CTL-1273 + CTL-1271 (roster keystone)

**Verdict: PASS** (draft; safe to mark ready after considering the one medium finding below). Verified the 5 PR-specific commits touch exactly the 10 declared files; ran the 5 suites under bun → **299 pass / 2 fail**, and the 2 failures are the pre-existing host-specific `getHostName (CTL-859)` tests — confirmed failing identically on unmodified `main` (this Mac's hostname is `Ryans-MBP-2.rozich`, so `getHostName` strips the first DNS label). Not attributable to this change.

### Required checks — all hold
- **(a) Live-fleet-safe.** `resolveClusterHosts` precedence is anchor → static → hosts-fallback → single-host. A daemon restart with a populated anchor reads the anchor; if the anchor is unreadable it falls through to the legacy `.catalyst/hosts.json`, and only an install with no anchor/static/hosts file lands on single-host. Today's live fleet (N=1, no anchor configured) takes the single-host branch with **no Linear call**.
- **(b) Fail-open.** Three layers, all tested: `readNodeRoster` returns `{}` on a `post` error (never throws); `readNodeNamesSync` returns `{ ok: false, names: [] }` on spawn/timeout/non-zero/unparseable; the resolver treats `ok:false` (and an empty `ok:true` roster) as "fall through," never as an empty fleet. `config.test.mjs` asserts the fall-through keeps the hosts.json roster — no mass-eviction path.
- **(c) Reader == writer.** `addHost`/`removeHost` default `anchor = getLivenessAnchorIssue()` and write via `registerNodeSync`/`deregisterNodeSync` to that anchor; `resolveClusterHosts` reads the same `getLivenessAnchorIssue()`. The `cli-cluster.test.mjs` anchor tests assert the write targets `anchorIssue: "CTL-1090"` and that hosts.json is untouched. Divergence is structurally impossible.
- **(d) hosts.json untouched.** `git diff` against the PR commits shows no change to any `hosts.json`; no CI guard added (the only "guard" strings are comments referencing the explicit follow-up + a test describe-name). Legacy path preserved as migration fallback.
- **(e) Single-host makes no Linear call.** The anchor read is gated on `getLivenessAnchorIssue()` being non-null; `config.test.mjs` proves `readNodeNames` is **never invoked** (real `called` flag) when no anchor is configured.
- **(f) Scope / reward-hacking / tests.** 10 files, exactly as declared. No `as any` / `@ts-ignore` / non-null `!` / `.only` / `forEach(async`. Scheduler dispatch, cluster-heartbeat, hrw, cluster-claim untouched. Tests are hermetic (inject `post` / `spawn` / `bootResolve`; no real network). Two existing `describe` blocks were renamed and pinned to `anchor: null` to keep exercising the legacy path — no assertion weakened. CTL-1271's three boot Gherkin scenarios are each asserted.

### Finding (medium) — per-tick anchor read has no cache
`getClusterHosts()` is called on the scheduler hot path (e.g. `scheduler.mjs:2773`, `:5522`) every tick. On a **multi-host deploy with the anchor configured**, each call now `spawnSync`s `node node-roster.mjs names <anchor>` doing a **synchronous Linear fetch** (15s timeout) inside the tick body — where the original design read a "cheap readFileSync" (the comment at `scheduler.mjs:2768-2769` still says so). The CTL-1273 note literally calls for "read node attachments from the anchor (**cache + boot-assert**)"; no cache layer was added. Live fleet is unaffected today (single-host = no call), and fail-open prevents a wrong roster, but a slow/hung Linear read would stall every tick up to 15s and multiply Linear read volume per the per-host 429 budgets. **Recommend a short-TTL cache for the anchor read before flipping any host to the anchor source** (can be a follow-up — it does not block the keystone).

### Finding (low) — `rename` not migrated to the anchor
`renameHost` still edits only `.catalyst/hosts.json` + Layer-2 `host.name`. On the anchor path a rename leaves the old `catalyst://node/<name>` attachment orphaned and the new name unenrolled, so the renamed daemon's self-name would be absent from the anchor roster. Out of CTL-1273's stated scope (add/remove), but worth a follow-up so the third cluster mutation also shares the single source.

Did not modify code or merge.

PR #2252, opened 2026-06-18
CTL-1274: Make the catalyst-cluster repo the roster source (retire Linear anchor source)

ryanrozich:
## CTL-1274 — roster source: Linear anchor (#2249) → catalyst-cluster repo

Swaps `resolveClusterHosts()` **source #1** from the Linear cluster anchor (built in #2249) onto the dedicated **catalyst-cluster GitHub repo** (`cluster.json.roster`, read via the existing CTL-1211 `readClusterConfig()` / `getClusterRepoDir()` machinery).

The pluggable resolver seam, the CTL-1271 boot assertion, fail-open, single-host default, and the hosts-fallback migration path all **stay**. Only source #1 changes; the Linear-anchor `node-roster` modules are removed.

### Changes
1. `resolveClusterHosts()` source #1 = `cluster-repo` (`readClusterConfig().roster`, schema-gated, fail-open). Precedence: **cluster-repo → static → hosts-fallback (legacy hosts.json) → single-host**.
2. Removed the Linear-anchor source: `node-roster.mjs`, `node-roster-sync.mjs` + their tests; reverted the `node-roster-sync` copy in `config-pino-fallback.test.mjs` (config.mjs no longer imports it).
3. CLI writer (`cli/cluster.mjs`): `add`/`remove`/`rename` edit `cluster.json.roster` in the cluster-repo clone + `git commit` + `git push` (reader == writer). Legacy `hosts.json` write path preserved when no cluster repo is present.
4. `cluster-sync` auto-pull: daemon calls `clusterSync` at boot + `pullClusterRepo` on a periodic env-overridable timer (fail-open). Reuses `cluster-sync.mjs`.
5. Boot assertion (CTL-1271): unchanged — now reports `source=cluster-repo`.

### Safety
- **Additive / non-destructive**: `.catalyst/hosts.json` is NOT deleted and NO CI guard is added — that is the post-verify cutover step (separate follow-up). With no cluster clone present, the resolver falls through to `hosts-fallback`, so landing + restart keeps the live fleet on `["mini","mini-2"]` — never silently single-host.
- Scheduler dispatch (CTL-1091) and liveness/heartbeat (CTL-1272) untouched.

Autonomous build — **NEEDS REVIEW**. Leaving CTL-1274 in Backlog; not marking ready / merging.


ryanrozich:
## Adversarial review — CTL-1274 (roster source → catalyst-cluster repo)

**Verdict: PASS** — contained swap, all hard claims verified, no scope creep, no reward-hacking. PR left as draft (correct — cutover is the follow-up).

### Hard claims verified
- **(a) Live-fleet-safe.** With no cluster clone present, `readClusterRepoRoster()` → `readClusterConfig()` returns null → falls through `cluster-repo → static → hosts-fallback`, and `.catalyst/hosts.json` (`["mini","mini-2"]`, untouched on the branch) wins. A restart keeps the live fleet multi-host, never silently single-host. Directly tested: `hosts-fallback (legacy .catalyst/hosts.json) when no cluster-repo + no static` and the too-new/empty fail-through tests land on `hosts-fallback`.
- **(b) Fail-open.** `readClusterRepoRoster` is schema-gated (`schemaCompat !== "too-new"`) and never throws; empty/malformed/too-new/missing → null → next source. Covered by `FAIL-OPEN: a too-new cluster schema falls through` and `an empty cluster roster falls through to static`.
- **(c) Reader==writer.** Both the reader (`config.mjs readClusterRepoRoster` → `readClusterConfig(getClusterRepoDir())`) and the CLI writer (`cli/cluster.mjs` `readClusterRoster`/`writeClusterRoster` → same `cluster.json` at `clusterDir`) key off `cluster.json.roster`. Writer commits + pushes; `writeClusterRoster` preserves anchorIssue/projects/schemaVersion (`{ ...cfg, roster }`), tested explicitly. Divergence is structurally precluded.
- **(d) Linear-anchor fully removed + pino-fallback green.** `node-roster.mjs`, `node-roster-sync.mjs` + both tests deleted; zero repo-wide references to `node-roster`/`readNodeNamesSync`/`registerNodeSync`/`deregisterNodeSync`. The `node-roster-sync` cpSync was reverted in `config-pino-fallback.test.mjs`; `config.mjs` load-time imports are back to node builtins + `config-schema.mjs` only. `config-pino-fallback.test.mjs`: **2 pass / 0 fail**.
- **(e) cluster-sync auto-pull fail-open.** Daemon calls `clusterSync()` at boot + `pullClusterRepo()` on an env-overridable timer (`EXECUTION_CORE_CLUSTER_SYNC_INTERVAL_MS`, default 5m). Both wrapped in belt-and-suspenders try/catch inside the boot try; timer `unref()`'d and cleared in `stopDaemon`; gated by `CATALYST_CLUSTER_SYNC=0` kill-switch. Tests: boot-throw doesn't abort, periodic-pull-throw keeps the timer firing (≥2 ticks), kill-switch skips both, stop clears the timer. `cluster-sync.mjs` reused unmodified.
- **(f) hosts.json untouched** (`["mini","mini-2"]`), **no CI guard added** — both correctly deferred to the cutover.

### Scope & quality
- PR diff (vs merge-base `fa9c2440`) touches **exactly** the 11 declared files (7 modified, 4 deleted). No scheduler/HRW/liveness changes.
- No `as any`/`@ts-ignore`/non-null `!`/`.only`/`.skip`/empty asserts.
- Tests: **287 pass / 2 fail** across the in-scope suites. The 2 failures are **pre-existing and unrelated** — `getHostName (CTL-859) > ...trailing .local stripped` and `...malformed Layer-2 file falls back`. They are byte-identical on `origin/main` (same `.replace(/\.local$/, "")` expectation), live in a block this PR did not touch, and fail only because this host's FQDN is `Ryans-MBP-2.rozich` (CTL-1252 collapses to the first DNS label `Ryans-MBP-2`, test expects `.local`-only stripping). Not a regression.

### Minor (non-blocking)
- Reader/writer activation asymmetry: the writer's `hasClusterRepo` gate keys on `cluster.json` *existing*, while the reader treats an empty-roster `cluster.json` as a miss (falls through). For an `add` this self-heals (write produces a non-empty roster the reader then sees); harmless. No end-to-end test wires the CLI writer output through the `config.mjs` resolver, but both sides demonstrably read the same file/key.

_Automated adversarial review. No code modified, not merged._

PR #2465, opened 2026-06-29
fix(dev): CTL-1382 — thin-launcher catalyst CLIs resolve their symlink before locating their driver

ryanrozich:
## CTL-1382 — `catalyst doctor` (and other thin launchers) broken via the installed symlink

**The bug you hit:** `catalyst doctor` → `catalyst doctor: missing ~/.catalyst/bin/execution-core/doctor.mjs`.

`catalyst-doctor` and `catalyst-otel-forward` derived `SCRIPT_DIR` from `BASH_SOURCE` **without resolving symlinks**, then looked for their `.mjs`/`.ts` driver relative to it. `install-cli.sh` installs each tool as a symlink in `~/.catalyst/bin`, so when invoked by name `SCRIPT_DIR` became the bin dir and the driver lookup missed.

## Fix
- **`catalyst-doctor`** + **`catalyst-otel-forward`**: add the canonical `readlink` resolution loop (shared with `catalyst` / `catalyst-install` / `catalyst-backup`) that normalizes each hop against the symlink's own dir, so **absolute and relative** symlink targets resolve to the real script dir.
- **`catalyst-why`**: hardened its existing loop to the same relative-target-safe idiom (it resolved symlinks but not relative targets).

`catalyst-why` already resolved symlinks (so it wasn't broken); the audit in CTL-1382 covered all three.

## Verification
`catalyst doctor` invoked through the `~/.catalyst/bin/catalyst-doctor` symlink now runs the full activation report (10 pass / 3 warn / 1 fail) instead of erroring. shellcheck clean (only an `SC1091` info on a sourced lib).

Found while fixing the related symlink-distribution gap (CTL-1381 — the updater doesn't re-run `install-cli` after a pull, so these newly-added tools were never symlinked on existing nodes).

### DIFF
--- a/plugins/dev/scripts/catalyst-doctor
+++ b/plugins/dev/scripts/catalyst-doctor
@@ -10,7 +10,18 @@
 # --dry-run accepted (documented as default read-only behavior, no separate code path).
 set -uo pipefail
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# Resolve symlinks before deriving SCRIPT_DIR — install-cli.sh installs this as a symlink in the bin
+# dir (~/.catalyst/bin), so a bare dirname would point there and the execution-core/doctor.mjs lookup
+# would miss ("catalyst doctor: missing <bin>/execution-core/doctor.mjs"). Normalize each hop against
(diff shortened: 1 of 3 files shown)


ryanrozich:
@codex review

PR #2500, opened 2026-06-30
fix(dev): CTL-1405 — sdkDispatch passes emitEvent in the wrong param → phase-turns telemetry never reached the event log

ryanrozich:
## Bug

CTL-1396 (#2484, merged) wired the daemon-injected unified-event-log `emitEvent` into the SDK dispatch path so `execution-core.sdk.phase-turns` (the turn-cap calibration signal) + `.overloaded`/`.auth.misconfigured` reach the JSONL event log / Loki. **It bound emitEvent to the wrong parameter, so it never took effect.**

`dispatch.mjs` (before):
```js
const launch = emitEvent ? (a) => runPhaseAgent({ ...a, emitEvent }) : runPhaseAgent;
```
`sdkRunPhaseAgent(input, { …, emitEvent = defaultEmitEvent, … })` reads `emitEvent` from its **second** (options) param. Spreading it into the **first** (input) object — which `sdkRunPhaseAgent` destructures as `{orchDir,ticket,phase,…}` and never reads emitEvent from — meant the injected emitEvent was ignored and `defaultEmitEvent` (stderr) was used.

**Caught by Codex** (P2 on #2484, post-merge) and **confirmed live 2026-06-30**: zero real `execution-core.sdk.phase-turns` events in the mini event log despite real SDK dispatches → the turn-cap calibration has had no data.

## Fix
```js
const launch = emitEvent ? (a) => runPhaseAgent(a, { emitEvent }) : runPhaseAgent;
```
The daemon-side binding (`daemon.mjs`) was always correct — only `sdkDispatch`'s forwarding was wrong.

## Test
The CTL-1396 test passed only because its stub read `args.emitEvent` (first-arg), matching the wrong assumption — it validated the wrong contract. The test now:
- mirrors the **real two-param** `sdkRunPhaseAgent` signature (`(input, opts)`, emitEvent in `opts`), and
- adds a guard asserting emitEvent is **not** smuggled into the input object.

Verified: the updated test **FAILS against the old `{ ...a, emitEvent }` shape**, passes against the fix. `dispatch` 74/0, `daemon` 122/0.

Resolves the unresolved Codex P2 on #2484. CTL-1398/executor=sdk validation is unaffected (that was real); this only restores the phase-turns/overloaded/auth telemetry wiring.

### DIFF
--- a/plugins/dev/scripts/execution-core/dispatch.mjs
+++ b/plugins/dev/scripts/execution-core/dispatch.mjs
@@ -222,7 +222,12 @@ export function defaultDispatch(
 // (other callers / unit tests) → sdkRunPhaseAgent keeps its dependency-free stderr
 // default, byte-identical to before.
 export function sdkDispatch(args, { runPhaseAgent = sdkRunPhaseAgent, emitEvent, ...seams } = {}) {
-  const launch = emitEvent ? (a) => runPhaseAgent({ ...a, emitEvent }) : runPhaseAgent;
+  // CTL-1405: sdkRunPhaseAgent reads `emitEvent` from its SECOND (options) param,
+  // NOT the first (input) object — so bind it there. The prior CTL-1396 shape
+  // `runPhaseAgent({ ...a, emitEvent })` spread it into the input object, where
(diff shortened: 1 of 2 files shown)


ryanrozich:
@codex review

PR #2512, opened 2026-07-01
feat(dev): CTL-1397 (4/n) — gate replica discovery on the writer heartbeat lock (close the quiet-feed linearis fallback)

ryanrozich:
## Why

The board-freeze fix (#2502 board-list→replica, #2509 empty-trust) only **half-worked** in production. Live verification after deploy:

- `eligible_source` = **~50% replica / ~50% linearis** during an active feed;
- **~100% linearis during a QUIET feed** (the reader MISSed 6/6 for CTL/Todo on mini).

### Root cause — the freshness gate false-positives during a quiet feed

`isReplicaFresh` gated on the db/`-wal` mtime. But `-wal` mtime only advances on an actual **apply**. During a quiet Linear feed (live writer, no issue updates) the `-wal` goes stale within the 5-min threshold even though the replica is **perfectly current** — so `eligible()` returns a MISS and discovery false-falls-through to `linearis issues list`, burning the shared quota + tripping the CTL-679 breaker **exactly when the board is unchanged**.

Observed live on mini: `-wal` **520s** stale while the writer was demonstrably alive — `.writer.lock` heartbeated **5s** ago and the cursor advancing (330766). The `.writer.lock` mtime advanced +10s over a 7s sample → it heartbeats.

## What

Gate on the cloud-sync writer's **heartbeat file `<db>.writer.lock`** (touched every few seconds regardless of data changes — already doctor's writer-liveness signal), not the update-driven `-wal`:

- lock **fresh** → writer alive → serve the (current) replica, even on a quiet feed;
- lock **present but stale** → writer died → do NOT serve (authoritative, even if a final apply left `-wal` fresh);
- lock **absent** → fall back to the db/`-wal` mtime (bootstrap / older writer).

Threshold unchanged (`CATALYST_LINEAR_REPLICA_STALE_MS`, default 5 min). Eligible-path only (`lookup`/`freshness`/`titles` already fail-open on error).

## Tests

`replica-read` **42 pass** — added **QUIET-FEED** (fresh lock + stale db/`-wal` → serves) and **DEAD-WRITER** (stale lock authoritative → undefined even with fresh db/`-wal`); existing stale/override tests exercise the lock-absent fallback.

## Verify after deploy

On mini/mini-2, `eligible_source` should go to **~100% replica** (with periodic linearis re-confirms per #2509's confirmed-empty window), and the CTL-679 breaker → quiet even through quiet-feed periods.

https://claude.ai/code/session_01Dfk5ZoEmBbh7SMamrCcsFW

### DIFF
--- a/plugins/dev/scripts/execution-core/replica-read.mjs
+++ b/plugins/dev/scripts/execution-core/replica-read.mjs
@@ -99,15 +99,34 @@ const RELATIONS_INVERSE_SELECT = `SELECT type, issue_identifier FROM relations W
 // cursor-present = seed complete; cursor-absent/empty = mid-reseed → don't serve.
 const SEED_COMPLETE_SELECT = `SELECT 1 FROM sync_meta WHERE key = 'cursor' AND value IS NOT NULL AND value <> '' LIMIT 1`;
 
-// The eligible freshness GATE — the SAME mtime-based writer-liveness proxy as
-// linear-cli.mjs's openFreshReplica. A dead writer stops touching the file →
-// stale mtime → fall through to linearis (correct answer, just un-accelerated).
-// WAL mode: appends land in the `-wal` sidecar; the main DB mtime only advances
-// on checkpoint, so take the freshest of the DB + a NON-EMPTY `-wal` (an empty
-// `-wal` is a reader artifact, never a writer-liveness signal — see the spoof-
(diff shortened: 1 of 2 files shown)


ryanrozich:
@codex review — small freshness-gate change: gate replica discovery on the writer heartbeat lock (.writer.lock) instead of the update-driven -wal, so a quiet feed doesn't false-fall-through to linearis. Root cause + live evidence in the PR body.

PR #2540, opened 2026-07-02
fix(dev): CTL-1420 follow-up — tier daemon Linear reads off the shared app-actor bucket

ryanrozich:
## What

The execution-core daemon's Linear **reads and writes both authenticate as the same "Catalyst Orchestrator" app-actor**, so every host draws **one shared 5000/hr bucket** (root-caused in the CTL-1420 burn incident). This PR is the **shared-bucket read-tiering quick-wins** follow-up (Ryan's go, fleet-channel #116) — three correctness-preserving cuts that trim redundant cold `linearis` spawns off that bucket. Separate PR from the heartbeat-400 fix (#2539).

## Changes

**1. Boot double-reconcile dedup** (`recovery.mjs`, `daemon.mjs`)
`recoverStartup` ran a full every-team `reconcileAll` **before** the read-replica reader is injected — so with `CATALYST_LINEAR_REPLICA=on` it fell to `linearis issues list` and drew the shared bucket — and `startMonitor` ran an **immediate replica-backed** reconcile right after: two boot polls. `recoverStartup` now takes `skipReconcile` (the daemon passes `true`); `startMonitor`'s replica-backed pass becomes the sole boot reconcile.
_Safe:_ nothing between `recover()` and `startMonitor` (`reconcileBoot`, `processApprovedResumes`) reads the eligible projection, and `report.routing` comes from the registry (`listProjects`), not the reconcile.

**2. `applyTriageStatus` pre-read tiering** (`linear-write.mjs`, `monitor.mjs`)
The Todo→Triage write-back's **pre-transition** read now threads `gateway` (CTL-823) + `replica` (CTL-1340), so a warm hit skips the cold read (the ticket was just admitted from the replica-served eligible set → high hit rate). The **post-write verify read-back stays live** — the lagging tiers would not yet reflect the just-written state and would false-report `verify-failed` on every triage.

**3. `classifyTicketResolution` replica tier** (`linear-query.mjs`, `daemon.mjs`)
The CTL-671 phantom-sweep existence probe gains a replica short-circuit mirroring the gateway one: a present replica row proves `"exists"` (the only verdict safe to serve without a live read). A **MISS can never yield `"not-found"`** — it falls through to gateway + live, so a lagging/absent replica can never quarantine a real ticket (fresh-before-quarantine holds).

## Deliberately NOT tiered (deviation from the handoff)

The `runTransition` / `applyPhaseStatus` **backward-write guard (CTL-758)** is left **cache-only**. That guard refuses to un-complete a Done ticket and depends on *fresh* state; a ≤60s-stale replica/gateway read on a cache-miss would re-open the CTL-758 regression for little burn benefit (its cache is warm via write-through).

## Deferred (separate follow-up, needs a decision)

The **per-host-key split** (route daemon reads onto each host's own `lin_api` key, off the shared app-actor bucket) is a genuine design decision — it collides with the CTL-679 read/write **breaker coupling** (one 429 pauses both). Not in this PR.

## Safety / testing

- All new params are **optional and default-off** → every existing caller is byte-identical.
- New tests: replica-tier `classifyTicketResolution` (HIT / MISS / precedence / never-not-found), `recoverStartup` `skipReconcile` (no poll, projection not written, routing intact), and the triage **pre-read-tiered vs verify-read-back-live** assertion.
- **Full execution-core suite baseline** (`env -u LINEAR_API_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN -u ANTHROPIC_API_KEY bun test`):
  - `origin/main`: 5764 pass / **34 fail**
  - this branch: 5773 pass / **34 fail**
  - → **+9 passing tests, 0 new failures.** The 34 are the documented pre-existing env-dependent failures (missing `@opentelemetry/sdk-trace-base`, pino shim, hostId/roster-dependent scheduler + `linearis fallback` + `janitor` tests) — identical on both.


ryanrozich:
@codex review

ryanrozich:
Superseded by the comprehensive reads→replica work. #2551 (CTL-1420 Stage-0) moves the dispatch-path reads to the local replica *with a currency gate* (confirm-null/trust-non-null), which properly addresses the stale-replica risk Codex flagged here; #2552/#2554 (CTL-863) cache the fence reads. Closing in favor of that vetted approach rather than merging a version with the known staleness. See thoughts/shared/plans/2026-07-02-dispatch-no-live-linear.md.

PR #2541, opened 2026-07-02
fix(dev): CTL-1420 follow-up — split daemon Linear READS onto the per-host key (v1: key split, shared breaker)

ryanrozich:
## What

The execution-core daemon authenticates **all** its Linear calls as the shared "Catalyst Orchestrator" app-actor (CTL-785), so every host's reads *and* writes draw **one shared 5000/hr bucket** — the root cause of the CTL-1420 burn. This routes the daemon's **READS** onto each host's own per-host `lin_api` key (its own 2500/hr bucket), leaving **WRITES** on the shared app-actor. Reads stop competing with writes and stop cross-contending across hosts. (Ryan approved the read/write key split.)

**This is v1 — the key split only.** The CTL-679 breaker stays **shared** (conservative: a 429 on either bucket still backs off both, same safety as today, and it leaves the CTL-1420 admission-freeze logic that keys on `breakerIsOpen` untouched). A follow-up (v2) can split the breaker per-bucket for storm-time isolation.

## Changes

- **Launcher (`catalyst-execution-core`)** — before CTL-785 overwrites `LINEAR_API_TOKEN` with the minted app-actor token, capture the inherited per-host key into `CATALYST_LINEAR_READ_KEY`. `set -u`-safe via `:-` defaults; only exported when a non-empty key was inherited (else reads fall back to the app-actor — byte-identical).
- **linearis reads (`linear-query.mjs` `rawExec`)** — when `CATALYST_LINEAR_READ_KEY` is set, spawn the `linearis` child with `LINEAR_API_TOKEN`/`LINEAR_API_KEY` overridden to the read key (covers both the capped and uncapped spawn branches via the shared `opts`).
- **curl GraphQL reads (`runBatchOnce` / `runDelegateOnce` / `runDelegateBatchOnce`)** — new `readAuthToken()` selects `CATALYST_LINEAR_READ_KEY` first, else `LINEAR_API_TOKEN` / `LINEAR_API_KEY`. Uses `||` (not `??`) so a defined-but-empty var never yields an empty `Authorization` header.

## Correctness

- The re-minter (`linear-remint.mjs`) only rewrites `LINEAR_API_TOKEN`/`LINEAR_API_KEY` on `process.env`, **never** `CATALYST_LINEAR_READ_KEY` — so a write-side app-actor remint can't leak into a read child; each read re-applies the read key.
- **WRITES are unchanged**: `linear-write` / `linear-transition.sh` / `linearis` read `LINEAR_API_TOKEN` from the inherited env, which the launcher still pins to the app-actor. The heartbeat write (`cluster-heartbeat.mjs`) likewise stays on it.
- When `CATALYST_LINEAR_READ_KEY` is unset (non-fleet host / mint failed), every path falls back to `LINEAR_API_TOKEN` → **byte-identical** to the pre-split behavior.
- Bun note: `spawnSync` with `env` omitted uses the process's *startup* env snapshot; the SET path passes an explicit `env`, so the read key applies. The launcher sets the env before daemon start, so production is correct.

## Testing

- `readAuthToken` selection (read-key-wins / fallback / empty-string / all-unset) and `rawExec` child-env injection (SET → child sees the read key; UNSET → identical to a plain omitted-env spawn).
- Launcher capture smoke-checked under `set -u` (per-host-token / per-host-key-only / neither).
- **Full execution-core suite**: `origin/main` 5764 pass / 34 fail → this branch 5771 pass / **34 fail** = **+7 passing, 0 new failures** (the 34 are the documented pre-existing env-dependent failures).

## Deploy

**NOT deployed.** This changes how the daemon authenticates to Linear fleet-wide, so it awaits an explicit deploy go (per the standing directive). Deploy = `catalyst-execution-core daemon restart` per host, then verify reads draw the per-host bucket.

## Follow-up (v2, not in this PR)

Split the CTL-679 breaker per-bucket (independent read/write breakers) so a read-bucket 429 doesn't pause writes and vice-versa. Deferred — re-points the admission and recovery breaker consumers onto the read/write instance respectively.


ryanrozich:
@codex review

ryanrozich:
Superseded by reads→replica (CTL-1420 Stage-0 #2551 + fence caches). A per-host read-key split is moot once the daemon's reads come from the local replica/event-log rather than live Linear — there's no shared read bucket to split off. Closing in favor of the durable no-live-Linear-reads direction. See thoughts/shared/plans/2026-07-02-dispatch-no-live-linear.md.

PR #2551, opened 2026-07-03
fix(dev): CTL-1420 — dispatch reads via replica (Stage 0: ownership→replica, eligible breaker-aware)

ryanrozich:
## Context
Stage 0 of severing the execution-core **dispatch path from live Linear** — the durable fix for the freeze class where the shared app-actor bucket saturates, the CTL-679 breaker opens, and dispatch (which read Linear live for ownership + eligibility) freezes. Design (adversarially vetted, 10-agent workflow): `thoughts/shared/plans/2026-07-02-dispatch-no-live-linear.md`.

**Root-cause note:** the fleet's live proxy capture named the dominant burner as `query ReadFence` (~80% of Linear traffic, CTL-863 fence guards) — owned separately (READ-SDK's ReadFence cache). **This PR is the complementary dispatch-read layer**, not the ReadFence fix.

## Changes (Stage 0 — keeps the live fallback; full elimination + event-log claims are Stage 1/2)
- **`replica-read.mjs`** — new `ownership(id)` reader (assignee+delegate), gated on writer-liveness + seed cursor **identically to `eligible()`**. `eligible()` now serves `project`/`label` filters inside the same snapshot (**D1**) instead of a live-linearis fall-through.
- **`linear-query.mjs`** — **A1 (confirm-null, trust-non-null):** `fetchTicketAssignee` consults `ownership()` first; a **non-null** delegate is actor-authoritative → trusted **Linear-free**; a **null** delegate falls through to the unchanged gateway/live confirm (no stomp on a stale null). **D2:** `runEligibleQuery` no longer spawns `linearis` into an **open** breaker on a replica miss — preserves the prior eligible set + emits `catalyst.alert.eligible_source_unavailable`.
- **`dispatch-alert.mjs`** (new) — one-shot best-effort `catalyst.alert.*` emitters to the event log.
- **`scheduler.mjs` / `monitor.mjs`** — thread the daemon's replica reader into the CTL-1174 gate.

## Design decision
**Rejects the breaker split** (an earlier idea): the daemon has a *single* shared app-actor token/bucket/breaker — there is nothing to split. Freeze-immunity comes from removing dispatch's *dependence* on the bucket (reads→replica), not from breaker surgery.

## Tests
replica-read **55**, linear-query **196**, monitor **137** pass; **zero net-new failures** in the full `execution-core/` suite; `node --check` clean. No breaker/claim files touched.

## Deploy
Merge only auto-reloads the broker; the exec-core change needs an explicit restart. **I'll defer that restart** until the fleet's ReadFence cache lands + unfreezes and the mini-2 audit-proxy window closes, then deploy + verify dispatch with a before/after proxy capture.

### DIFF
--- a/plugins/dev/scripts/execution-core/dispatch-alert.mjs
+++ b/plugins/dev/scripts/execution-core/dispatch-alert.mjs
@@ -1,0 +1,118 @@
+// dispatch-alert.mjs — Stage 0 of "dispatch severed from live Linear"
+// (2026-07-02-dispatch-no-live-linear.md). One one-shot, best-effort
+// `catalyst.alert.*` emitter for the Stage-0 dispatch-hardening signal:
+//
+//   catalyst.alert.eligible_source_unavailable (WARN) — linear-query.mjs
+//     runEligibleQuery(): the eligible poll MISSED the replica AND the CTL-679
+//     breaker is OPEN, so we preserve the prior set WITHOUT spawning linearis into
+//     the open breaker (D2). The genuine no-source residual (fresh boot, no replica,
+//     no prior set) is now LOUD instead of a silent whole-window freeze.
+//
+// Emit path mirrors fleet-freeze-alert.mjs (the execution-core alert precedent):
+// buildCatalystResource envelope + appendFileSync to the canonical event log,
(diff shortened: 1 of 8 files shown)


ryanrozich:
@codex review

PR #2552, opened 2026-07-03
fix(dev): CTL-863 — interim in-process TTL cache for fence ReadFence to unfreeze the shared Linear bucket

ryanrozich:
## Summary

- The CTL-863 fence guards (`fenceGuard`, called before every daemon-side external-write site — ~11 sites across `scheduler.mjs`, `recovery.mjs`, `stale-pr-rescue-timer.mjs`) hit Linear's `query ReadFence` (an attachment read on the ticket's issue) before each write. Live-proxy-confirmed at **~5,000/hr — 62% of ALL Linear traffic** on the shared app-actor bucket, saturating it and tripping the **CTL-679 rate-limit breaker open**, which **freezes fleet dispatch entirely** (zero dispatches for hours).
- Adds `fenceCheckSyncCached` in `cluster-claim-sync.mjs`: an **in-process TTL cache** (default **45s**, override via `CATALYST_FENCE_READ_CACHE_MS`, `0` disables) keyed by `` `${ticket}::${generation}` `` — wired in as `fence-guard.mjs`'s default `check`. The underlying fence only changes on the ~**2min heartbeat cadence** (`cluster-heartbeat.mjs`), so a 45s cache can never observe a staler fence than genuinely exists at write time — safe by construction, not a race.
- Only a **determinate** read is cached (`current:true`, or a confirmed-stale `current:false,stale:true`). The **indeterminate/error** bucket (spawn error, timeout, other exit) is **never cached** — `fenceGuard` is fail-closed on it, so caching a transient hiccup would extend a false "not current" (write-suppressing) verdict for the full TTL instead of retrying on the very next call.
- `fenceGuard`'s decision logic and fail-closed semantics are **untouched** — this only reduces read volume, never changes what a check means. The `UpsertFence` write path is untouched (read-only caching).
- **This is the INTERIM unfreeze.** The durable fix replaces the read-per-check fence with an event-log-derived one: `thoughts/shared/plans/2026-07-03-fence-to-eventlog.md`.

**Expected effect**: CTL-679 breaker-opens → 0, dispatch resumes fleet-wide.

## Files changed

- `plugins/dev/scripts/execution-core/cluster-claim-sync.mjs` — new `fenceCheckSyncCached` + `clearFenceReadCache` (test seam), wrapping the existing (unchanged) `fenceCheckSync`.
- `plugins/dev/scripts/execution-core/fence-guard.mjs` — default `check` seam switched from `fenceCheckSync` to `fenceCheckSyncCached`; doc comment updated.
- `plugins/dev/scripts/execution-core/cluster-claim-sync.test.mjs` — new test coverage for the cache: TTL hit/miss, TTL expiry, env-disable, error-not-cached, generation-scoped keys, and argv passthrough on a cache miss.

## Test plan

- [x] `env -u CLAUDE_CODE_OAUTH_TOKEN -u ANTHROPIC_API_KEY -u LINEAR_API_TOKEN bun test cluster-claim-sync.test.mjs fence-guard.test.mjs fence-guard-integration.test.mjs cluster-claim.test.mjs` → **77 pass / 0 fail**
- [x] `node --check` on all three changed source files
- [x] Confirmed the 11 pre-existing failures in `scheduler.test.mjs`/`recovery.test.mjs`/`stale-pr-rescue-timer.test.mjs` are unchanged before/after this diff (824 pass / 11 fail both on `main` and on this branch) — this change introduces zero regressions.

### DIFF
--- a/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
@@ -124,3 +124,82 @@ export function fenceCheckSync(
     return { current: false, stale: false };
   }
 }
+
+// ─── in-process TTL cache around the fence read (CTL-863 fleet-unfreeze, urgent interim) ──
+//
+// The CTL-863 fence guards (fenceGuard, fence-guard.mjs) call fenceCheckSync
+// before EVERY external-write site — ~11 call sites across scheduler.mjs,
+// recovery.mjs, and stale-pr-rescue-timer.mjs. Each call spawns a FRESH `node
+// cluster-claim.mjs fence-check <ticket> <gen>` subprocess (a new process, so
+// caching INSIDE cluster-claim.mjs would be cold every time) that issues
+// Linear's `query ReadFence` (an attachment read). Live-proxy-confirmed at
+// ~5,000/hr — 62% of ALL Linear traffic on the shared app-actor bucket —
+// saturating it and tripping the CTL-679 rate-limit breaker open, which
+// freezes fleet dispatch entirely.
+//
+// fenceCheckSyncCached wraps fenceCheckSync with an in-process TTL cache that
+// lives in THIS module (imported once by the long-running daemon process —
+// scheduler.mjs / recovery.mjs / stale-pr-rescue-timer.mjs — so the Map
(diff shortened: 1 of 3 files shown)


ryanrozich:
@codex review

PR #2554, opened 2026-07-03
fix(dev): CTL-863 — cache the fence entourage (ResolveIssueId anchor-UUID + GetIssueByIdentifier) to fully clear the shared bucket

ryanrozich:
## Summary

Urgent follow-up to #2552. That PR added a 45s TTL cache for the `ReadFence`
read (82% of Linear traffic on the shared app-actor bucket) and materially
cut the burn (breaker trips 18/3m → 1-2/3m per OTEL), but left the
**uncached entourage** still keeping the bucket at the 429 edge:

- `query ResolveIssueId` — resolves an identifier → issue UUID. Issued on
  every claim (`cluster-claim.mjs::writeClaim`) and every ~2min heartbeat
  publish (`cluster-heartbeat.mjs::publishHeartbeat`) — the anchor/ticket
  UUID never changes once resolved, so this is the highest-value, safest
  cache: **permanent, no TTL**.
- The peer-heartbeat "get issue by identifier" read
  (`cluster-heartbeat.mjs::readPeerHeartbeats`, consumed via
  `readPeerHeartbeatsSync` in `recovery.mjs`'s dead-host detection) — a
  read-only liveness check, safe to cache with the **same 45s TTL** pattern
  as the ReadFence cache (dead-host detection already tolerates a 10-min
  staleness grace).

## Changes

- **`cluster-claim.mjs` / `cluster-claim-sync.mjs`**: `writeClaim` /
  `claimTicket` / `claimDispatchSync` gain an optional pre-resolved
  `issueId` override. A new `resolve-issue-id` CLI subcommand +
  `resolveIssueIdSyncCached` (permanent in-process `Map`, env
  `CATALYST_ANCHOR_UUID_CACHE`, "0" disables) let the daemon-side sync
  wrapper resolve a ticket's UUID once and skip the redundant
  `ResolveIssueId` round-trip on every subsequent claim. The CAS reads
  (`readClaim`, inside `claimTicket`) are deliberately **left uncached** —
  they are the actual fencing correctness check, not an immutable mapping,
  so caching them would risk a false win/lose.
- **`cluster-heartbeat.mjs` / `cluster-heartbeat-sync.mjs`**: the same
  permanent-cache pattern for the anchor issue (`publishHeartbeat` gains an
  `issueId` override + a `resolve-anchor` subcommand), plus a 45s TTL cache
  (`readPeerHeartbeatsSyncCached`, reusing `CATALYST_FENCE_READ_CACHE_MS`)
  around the peer-liveness read.
- **`recovery.mjs`**: wires `readPeerHeartbeatsSyncCached` into
  `readClusterHeartbeats` / `defaultOwnedTicketsForHost`.

Both caches only ever cache a **determinate success** — never an
error/indeterminate result — matching #2552's Map+TTL+env+test-seam style
exactly. Fence/claim write semantics are untouched; this only avoids
redundant Linear round-trips.

## Test plan

- [x] `node --check` on all 9 touched files
- [x] `env -u CLAUDE_CODE_OAUTH_TOKEN -u ANTHROPIC_API_KEY -u LINEAR_API_TOKEN bun test` on the 4 directly-touched test files: **141 pass / 0 fail**
- [x] Same for `fence-guard.test.mjs` + `fence-guard-integration.test.mjs` (consumers of `cluster-claim-sync.mjs`): pass
- [x] `recovery.test.mjs` (consumer of the new heartbeat cache): pass in isolation. Note: this file has a **pre-existing, unrelated** flakiness — `recoverStartup`'s "rebuilds routing state" test makes a real unauthenticated network call to `api.linear.app` (no `delegateExec` mock) via `runEligibleQuery`'s delegate-batch enrichment; enough repeated real hits from this sandbox trip Linear's live rate limiter, which opens the shared `linearBreaker` singleton for 60s and cascades into unrelated `reclaimDeadWorkIfPossible` test failures. **Reproduced this with zero code changes** (clean checkout, repeated runs) — confirmed unrelated to this diff.

### DIFF
--- a/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
@@ -36,12 +36,49 @@ const CLAIM_TIMEOUT_MS = Number(process.env.EXECUTION_CORE_CLAIM_TIMEOUT_MS) || 
 // via the env passthrough at the spawn call below. When set, overrides the
 // 300_000 ms (5 min) default stale-claim preemption threshold (CTL-1297).
 
-// claimDispatchSync — soft-CAS claim `ticket` for `hostName` at `phase`,
-// synchronously. Returns { won, generation }. won:false on any failure
-// (fail-closed). `spawn`/`nodeBin`/`cli`/`env`/`timeout` are injectable so the
-// unit tests never spawn a real process.
-export function claimDispatchSync(
-  { ticket, hostName, phase },
+// ─── permanent in-process cache: ticket identifier → issue UUID (CTL-863 fleet-unfreeze, entourage follow-up to #2552) ──
+//
+// #2552 cached the ReadFence read (82% of the fence traffic) below via
+// fenceCheckSyncCached, but left the "entourage" queries uncached: every claim
+// resolves the ticket's identifier → UUID via `query ResolveIssueId` inside writeClaim,
+// even though a ticket's issue UUID can never change once Linear assigns it. Unlike the
+// ReadFence read (a TTL cache, because the underlying claim/fence state genuinely
+// changes on a cadence), this is safe to cache PERMANENTLY — there is no staleness
+// window to reason about.
+//
+// resolveIssueId is bundled inside the `claim` CLI subcommand (one subprocess call does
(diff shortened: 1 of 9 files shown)


ryanrozich:
@codex review

PR #2556, opened 2026-07-03
feat(dev): CTL-1417 — worktree-removal self-protection guard

ryanrozich:
## CTL-1417 — worktree-removal self-protection guard

Closes the env-leak vector where a shell `git worktree remove --force` could delete the operator's own worktree (or one a live process is inside). Ports the CTL-791 `worktree-safety.mjs` cwd/lsof gate to the shell sites the Node reaper never covered, wires it into every `--force` removal call, and hardens the suspect test suites against the `$HOME` leak that made the incident reproducible.

### Phase 1 — the guard (`lib/worktree-remove-guard.sh`)

`assert_worktree_removal_safe <target>` refuses a removal when:

- `$PWD` is at-or-under the target (removing the ground under your own feet), or
- any process holds a live handle under the target (`lsof`).

Fail-closed: if the `lsof` probe cannot run, the removal is refused. The `lsof` binary is injectable via `WT_GUARD_LSOF` for deterministic unit tests. Ships the lib + a unit suite covering all branches (incl. fail-closed) + a `run-tests.sh` discovery wrapper so `make test` covers it.

### Phase 2 — route every `--force` removal through the guard

`assert_worktree_removal_safe` is now wired into every shell `git worktree remove --force` site:

- **`phase-agent-dispatch` CTL-707 recreate** — `cd` out to the repo root *before* the removal, then guard; on refusal fall through to park `stalled` rather than destroying. The cd-out keeps the guard non-breaking in the legitimate recreate (nothing else under the tree → `lsof` rc=1 → allowed) yet firing under an env-leak (a runner still holding the tree → refuse).
- **`create-worktree.sh`** — both rollback sites skip the force-remove on refusal, leaving the tree for the reaper.
- **`orphan-sweep.sh`** — SAFE + SALVAGE_UNPUSHED skip (`activeSkipped`) on refusal, so the hourly LaunchAgent can never yank a tree an operator / `make test` is inside.
- **`phase-teardown`** — added as a defense-in-depth belt beside the existing primary-worktree + presweep guards.

### Phase 3 — harden suspect suites' hermeticity

Give `phase-agent-dispatch.test.sh` and `phase-teardown-e2e.test.sh` the same isolation floor as the `orphan-sweep.test.sh` gold standard: a scratch `HOME` and `GIT_CONFIG_GLOBAL`/`SYSTEM=/dev/null`, plus a Red canary that fails loudly if the isolation is ever not in effect. This closes the env-leak vector at the source — no future code path that computes `$HOME/catalyst/wt/...` (e.g. the CTL-707 recreate's `create-worktree.sh`) can resolve a removal target to the operator's real worktree. The dispatch suite also pins `CATALYST_RECREATE_WORKTREE_DIR` under scratch as a belt.

### Tests

- New unit suite for the guard (all branches incl. fail-closed).
- Dispatch **T38c** (recreate refused → parks, tree survives) with a deterministic `WT_GUARD_LSOF` seam.
- Orphan-sweep **T24g** (guard-refused SAFE tree skipped, survives).
- New `create-worktree-guard` suite (structural wiring + gated removal decision).
- Existing SAFE-removal + teardown suites stay green.

Refs: CTL-1417

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-guard.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-guard.test.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# CTL-1417: create-worktree.sh rollback removals must route through the shared
+# self-protection guard (assert_worktree_removal_safe). When the guard reports a
+# live foreign handle under the worktree, the `--force` rollback removal is
+# SKIPPED (tree left for the reaper) instead of yanking an in-use tree. When the
+# guard is clear, the rollback removes exactly as before.
+#
+# Drives rollback site 1 (auto-detected `make setup` failure) by committing a
+# Makefile whose `setup` target exits non-zero. lsof is stubbed via WT_GUARD_LSOF
+# so the guard verdict is deterministic.
+#
+# Run: bash plugins/dev/scripts/__tests__/create-worktree-guard.test.sh
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../../.." && pwd)"
+CREATE_WT="${REPO_ROOT}/plugins/dev/scripts/create-worktree.sh"
+
+FAILURES=0
+PASSES=0
+fail() {
+	FAILURES=$((FAILURES + 1))
+	echo "  FAIL: $1"
+}
+pass() {
+	PASSES=$((PASSES + 1))
+	echo "  PASS: $1"
+}
+assert_eq() {
+	if [[ $1 == "$2" ]]; then pass "$3"; else fail "$3 — expected '$1', got '$2'"; fi
+}
+
+SCRATCH="$(mktemp -d -t cwt-guard-XXXXXX)"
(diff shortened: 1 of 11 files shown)


ryanrozich:
@codex review

ryanrozich:
Addressed all automated-review findings across two commits (recovery-pass, CTL-1417):

**9b093c4b** — P1: exempt the teardown worker's own process tree from the `lsof +D` guard (was detecting itself → would leave every merged worktree behind); P2: 10s cap on the recursive lsof; P2: fail-closed on an inconclusive lsof (stderr diagnostic); P2: count guard-refused salvage trees as `activeSkipped`.

**d3727354** — P1: guard-absence now fails **closed** at every `git worktree remove --force` site (single `_removal_guard_ok` predicate across orphan-sweep SAFE+salvage, create-worktree ×2 rollbacks, phase-agent-dispatch recreate); P2: dedicated `worktree_live_handle_guard_refused` stall reason so unstuck-sweep doesn't mis-route it to force-push; P2: let the negated-removal test assertion actually fail; fixed the T24g mock to emit real `lsof -F pn` format.

Suites green: worktree-remove-guard 11/0, orphan-sweep 206/0, create-worktree-guard 6/0; CI execution-core-unit-tests + Cloudflare Pages pass. Resolving these threads.

PR #2575, opened 2026-07-07
feat(dev): CTL-1420 (#17) — move cross-host liveness off Linear → event log + Loki (PR1a: heartbeat carries in-flight tickets)

ryanrozich:
## What & why

Part of moving **process-state (heartbeats) off Linear** (Ryan's directive, 2026-07-07) — the execution-core daemon publishes a `catalyst://heartbeat/<host>` Linear **attachment** every 120s (the ~120/hr write that flaps the CTL-679 breaker on the shared app-actor bucket). That attachment is the **only** cross-host transport for a peer's `last_seen` *and* `in_flight_tickets`.

The replacement is the unified event log → **Loki** (already central + cross-host; liveness is fail-open, so no new store/mesh needed — see dossier). `node.heartbeat` already reaches Loki with host + freshness, but the in-flight list lived in `body.payload`, which `otel-forward` **strips** (`otlp.ts:51` forwards only `attributes`).

## This PR (PR1a — additive telemetry, no behavior change)

Promotes the in-flight ticket list to top-level **attributes** so it survives to Loki as **structured metadata** (not an indexed stream label — collector cardinality rule):
- `catalyst.node.in_flight_tickets` (comma-joined) + `catalyst.node.in_flight_count`
- Daemon supplies the list via `localInFlightTickets(self)` (the same local signal-scan the publisher uses; now exported)
- Fail-safe: non-array/empty → `[]`. Single-host unchanged.

**Nothing reads this yet** — safe to deploy standalone; it de-risks PR1b by getting the data flowing to Loki first.

## Follow-ups (separate PRs)
- **PR1b** — `loki-liveness.mjs` fail-open reader + rewire `readClusterHeartbeats`/`defaultOwnedTicketsForHost` to Loki behind `CATALYST_LIVENESS_READ_SOURCE` (default `loki`, escape-hatch `linear`) + retire the Linear heartbeat publish (keep the Linear-free fence re-emit). This is the burn win.
- **PR2** — claim ownership visibility (`worker:<host>` label; gated on a Linear label-API write-test — single-select server-enforcement is unproven).

## Verification
- `heartbeat-event.test.mjs` — 4 new tests pass (attributes carry the list + count; empty default; dirty-input filtering; `startHeartbeat` forwards the seam). One unrelated pre-existing failure (`host.name defaults to os.hostname()`) fails identically on the base — it asserts the raw machine hostname but this host has a Layer-2 config name.
- `cluster-heartbeat-publisher.test.mjs` — 19/0 (export change clean). `daemon.mjs` builds clean.

Design: `thoughts/shared/research/2026-07-07_liveness-off-linear-dossier.md` · `thoughts/shared/plans/2026-07-07-liveness-off-linear-impl.md`

Deploy coordination via CTL-DELEGATE (channel `ctl-delegate-operational-unblock`).

### DIFF
--- a/plugins/dev/scripts/execution-core/cluster-heartbeat-publisher.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-heartbeat-publisher.mjs
@@ -64,7 +64,7 @@ function readLocalMaxParallel(orchDir) {
 // in defaultOwnedTicketsForHost (recovery.mjs) — factored here so the publisher
 // and the recovery fallback share the logic without importing recovery.mjs (which
 // would create a circular dependency once recovery imports this module's outputs).
-function localInFlightTickets(hostName, { orchDir } = {}) {
+export function localInFlightTickets(hostName, { orchDir } = {}) {
   if (!orchDir) return [];
   try {
     const signals = readWorkerSignals(orchDir);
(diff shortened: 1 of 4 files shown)


ryanrozich:
@codex review

PR #2604, opened 2026-07-09
feat(dev): CTL-1420 (#17) — read cross-host liveness from Loki + retire the Linear heartbeat publish (PR1b)

ryanrozich:
## What & why (the burn win)

Flips cross-host **dead-host detection** off the Linear heartbeat attachment — the ~120/hr write that flaps the CTL-679 breaker — onto the unified event log read via **Loki**, which is already the central place every host pushes `node.heartbeat` to (so it IS the cross-host transport; no mesh, no new store). LIVENESS ONLY — claim/fence CAS stays on its arbiter (Loki is append-only). Full design: `thoughts/shared/plans/2026-07-07-liveness-off-linear-impl.md`.

Builds on **PR1a #2575** (heartbeat carries `in_flight_tickets` to Loki). **Deploy PR1a first** (or both together), then flip the env — see rollout.

## Changes
- **`loki-liveness.mjs`** — fail-open Loki reader: newest `node.heartbeat` per host → `{host:{last_seen,in_flight_tickets}}`, defensive across the structured-metadata and stream-label shapes; `nsToMs` avoids Number-overflow on ns timestamps. `read` CLI for the sync bridge.
- **`loki-liveness-sync.mjs`** — spawnSync bridge (recovery.mjs is synchronous) + 20s TTL cache, mirroring `cluster-heartbeat-sync.mjs`. Fail-open → `{}`.
- **`config`** — `getLivenessReadSource()` (env `CATALYST_LIVENESS_READ_SOURCE`, default **`linear`** — safe rollout) + `getLokiQueryUrl()` (explicit env, else OTLP endpoint port-swapped to `:3100`, else `null` → fail-open).
- **`recovery.mjs`** — `readClusterHeartbeats` / `defaultOwnedTicketsForHost` read peers via a source-aware seam; the multi-host merge gate is `peerLivenessConfigured()` (loki: resolvable Loki URL; linear: anchor, unchanged). Fail-open: Loki down → peers absent → `deadHosts` treats them alive → **no false reclaim**.
- **`cluster-heartbeat-publisher.mjs`** — in loki mode the Linear anchor publish is skipped (burn retired); the Linear-free #2553 fence re-emit still runs.

## Rollout (safe, revertible)
[...]

PR #2606, opened 2026-07-09
fix(dev): CTL-1420 (#17) follow-up — loki-liveness reader: newest-across-streams + two-query tickets

ryanrozich:
## Two live-verification bugs in the merged loki-liveness reader (PR1b #2604)

Found while **deploying PR1b** to the fleet — I reverted both minis to `linear` (known-good) while fixing. This is why deploying and *observing* matters: unit tests passed on synthetic single-stream responses, but **real Loki behaves differently**.

### 1. Stale peer → false-reclaim risk (the serious one)
Real Loki returns **one single-value stream per heartbeat** (the changing `catalyst_node_in_flight_count` label splits every beat into its own stream). `parseLokiLivenessResponse` computed "newest" **per stream** and overwrote `out[host]` each stream, so the **last-listed (arbitrary, often stale) stream won** → a live peer read **~20 min stale** → past the 10-min grace → the daemon would false-flag it dead and reclaim its in-flight work. **Fix: track newest ts per host ACROSS all streams.**

### 2. Empty tickets
Loki only surfaces the `in_flight_tickets` structured metadata when a query **references** it, so the single query returned `in_flight_tickets:[]` for every host. **Fix: two fail-open queries** — A (liveness, robust, all hosts) + B (tickets enrichment via `| catalyst_node_in_flight_tickets=~`.+``, best-effort). A B-failure leaves ownership to the local/board-sweep backstop; liveness (A) is unaffected.

## Verification
- Regression tests: multi-stream-per-host newest-across-streams (**order-independent**) + two-query enrichment + B-failure-doesn't-break-liveness. loki 15, loki-sync 9, config 6 — all pass.
- **Live**: the reader now returns fresh `last_seen` (mini 30s / mini-2 28s, was 20min stale) + mini's 6 in-flight tickets.

## Rollout
Fleet is on `linear` (reverted) until this merges + deploys, then I re-flip `CATALYST_LIVENESS_READ_SOURCE=loki` and re-verify no false-reclaim.

### DIFF
--- a/plugins/dev/scripts/execution-core/loki-liveness.mjs
+++ b/plugins/dev/scripts/execution-core/loki-liveness.mjs
@@ -54,35 +54,64 @@ export function parseLokiLivenessResponse(body) {
   const out = {};
   const results = body?.data?.result;
   if (!Array.isArray(results)) return out;
+  // Track the newest ts per host ACROSS ALL streams. Real Loki returns ONE
+  // single-value stream PER heartbeat line (the changing catalyst_node_in_flight_count
+  // label splits every beat into its own stream), so "newest per host" MUST be a
+  // cross-stream reduction — a per-stream `newest` that overwrites out[host] each
+  // stream would let the LAST-listed stream win (a stale beat), which is exactly the
+  // bug that made a live peer read ~20 min stale and risked a false reclaim.
+  const newestMs = {};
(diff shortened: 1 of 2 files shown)


ryanrozich:
@codex review

PR #2636, opened 2026-07-14
fix(dev): CTL-764 follow-up — round-5: thenable-aware removal confirmation

ryanrozich:
Codex round-5 on the merged #2632 caught a real bug in the round-4 fix: production `removeLabel` is **async**, so the synchronous `res?.removed !== false` inspected a Promise — `.removed` read `undefined` and every removal **false-confirmed**, making the failed-removal suppression a no-op in prod (sync test stubs hid it).

Fix: the `onRemoveResult` seam is thenable-aware, and the admission clear emission moves **inside** the callback — it fires when the write resolves (`{removed:true}` → emit, `{removed:false}`/rejection → suppress; post-tick in prod, synchronous for sync stubs). `recordTransition`'s only-on-change guard dedupes the rare double-callback from a ticket wearing two stale held labels.

Tests: 3 new async-stub cases proving the prod shape (resolve-false suppresses, resolve-true emits with proven prior, rejection suppresses fail-open); full finding/converge set 27/0; adjacent suites 215/0.


ryanrozich:
@codex review

PR #2685, opened 2026-07-21
fix(dev): CTL-1497 — repair thoughts/shared on the worktree reuse path; guards test -L not -d

ryanrozich:
## Problem

The daemon always creates worktrees with `--reuse-existing` (`worktree.mjs:29`). In `create-worktree.sh`, the `--reuse-existing` block short-circuits with `exit 0` **before the thoughts-setup block ever runs**, so a worktree first created with a broken `thoughts/shared` (a plain directory instead of the humanlayer symlink — from an interrupted/failed init or a pre-existing dir) is **never repaired on any later dispatch**. Compounding it, the post-setup guards tested `-d`, not `-L`, so a plain-dir `thoughts/shared` *passed* — the guards only caught total absence, never the wrong-type state.

**Impact:** thoughts written into such a worktree (handoffs, research, plans) land in a local non-symlinked dir and never sync to the shared thoughts repo — silently stranded. Observed live: the CTC-263 dev-view handoff (catalyst-cloud) was written into a reused worktree whose `thoughts/shared` was a plain dir; it never synced, so `/resume-handoff` and `humanlayer thoughts sync` from the main checkout couldn't find it (recovered by hand).

This is the residual [CTL-513](https://linear.app/coalesce-labs/issue/CTL-513) did not close.

## Fix

- **Reuse path:** before the reuse-path `exit 0`, if `thoughts/shared` is not a symlink, repair it via the existing `catalyst-thoughts.sh init-or-repair` primitive (which moves a plain-dir aside and rebuilds the symlinks). Non-fatal on failure — prints the manual recovery command. `SCRIPT_DIR` isn't defined this early, so it's computed inline.
- **Harden guards:** both `[ ! -d "thoughts/shared" ]` checks → `[ ! -L "thoughts/shared" ]`, so a plain dir is caught, not just absence.

`bash -n` clean. No behavior change on the happy path (a valid symlink passes `-L`).

## Follow-up (same ticket)
Extend `plugins/dev/scripts/__tests__/create-worktree-thoughts-check.test.sh` with a case exercising the `--reuse-existing` path against a pre-existing plain-dir `thoughts/shared` (today only the CREATE path is covered). Optionally add a per-worktree `-L thoughts/shared` check to `execution-core/doctor.mjs` `checkThoughts()`.

Closes CTL-1497.

### DIFF
--- a/plugins/dev/scripts/create-worktree.sh
+++ b/plugins/dev/scripts/create-worktree.sh
@@ -139,24 +139,34 @@ if [ -d "$WORKTREE_PATH" ]; then
 		# CTL-1497: the reuse path short-circuits BEFORE the setup block below, so a worktree first
 		# created with a broken thoughts/shared — a plain directory, OR a dangling symlink whose target is
 		# gone — is never repaired on later dispatches, and thoughts written there strand and never sync.
-		# A HEALTHY thoughts/shared is a symlink that resolves to a directory (-L AND -d); repair anything
-		# else. init-or-repair refuses to clobber an existing plain-dir/dangling thoughts/ (it will not risk
-		# data loss), so move it aside first — stranded content is preserved under .orphaned-* — then
-		# rebuild. If the repair does not leave a healthy symlink, FAIL LOUD: never report a successful
-		# reuse of a worktree that would still strand thoughts.
+		# A HEALTHY thoughts/shared is a symlink that resolves to a directory (-L AND -d).
 		if [ ! -L "$WORKTREE_PATH/thoughts/shared" ] || [ ! -d "$WORKTREE_PATH/thoughts/shared" ]; then
-			_CW_SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-			echo -e "${YELLOW}  ⚠️  thoughts/shared is not a healthy symlink — repairing (CTL-1497)${NC}"
-			if [ -e "$WORKTREE_PATH/thoughts" ] || [ -L "$WORKTREE_PATH/thoughts" ]; then
-				mv "$WORKTREE_PATH/thoughts" "$WORKTREE_PATH/thoughts.orphaned-$(date +%Y%m%d-%H%M%S)" \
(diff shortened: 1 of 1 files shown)


ryanrozich:
Both findings were real bugs in the first cut — fixed in `3d7e55db`, and I verified the corrected logic against all five `thoughts/shared` states (healthy / plain-dir / dangling / absent / repair-fails):

**1. Do not report success when repair failed** — `init-or-repair` deliberately refuses to clobber a plain-dir/dangling `thoughts/` (it won't risk data loss), so it returns non-zero for the exact case this targets. The reuse path now (a) moves the broken `thoughts/` aside first — stranded content preserved under `thoughts.orphaned-*` — then rebuilds, so the plain-dir case actually heals; and (b) re-verifies a healthy symlink afterward and **exits 65** if it isn't, instead of `|| echo … ; exit 0`. A worktree that would still strand thoughts now fails loud.

**2. Reject dangling thoughts symlinks** — a healthy `thoughts/shared` is a symlink that *resolves to a directory*. All three guards (reuse path + both setup-block checks) now require `-L` **and** `-d`, so a dangling symlink (target gone) is caught, not just a plain dir or total absence.

Test matrix (stubbed `init-or-repair`): healthy→skip; plain-dir→heals + content preserved; dangling→heals; absent→heals; repair-fails→exit 65 (no false success).

ryanrozich:
Good catch — fixed in `57188c9`. The reuse-path repair (and its exit-65 hard-fail) is now gated on **shared thoughts being configured for the project**: it resolves the thoughts profile exactly as the setup block does — `.catalyst.thoughts.profile` from config, else `humanlayer thoughts status` — and only repairs when a profile resolves. An unconfigured project (no profile / HumanLayer absent) has no `thoughts/shared` by design, so the block is skipped entirely and reuse proceeds; phases 2–9 are unaffected.

Verified: configured+broken→heals; **unconfigured+plain-dir→skip (exit 0)**; **unconfigured+absent→skip (exit 0)**; configured+healthy→skip.

PR #2688, opened 2026-07-22
test(dev): CTL-1497 — reuse-path thoughts-repair regression tests

ryanrozich:
Adds the follow-up regression coverage noted in #2685 for the CTL-1497 fix — exercising the `--reuse-existing` path (the one the daemon always takes) that the fix repairs.

**Cases 8–10** (build a *pre-existing* worktree with a broken `thoughts/shared`, then run the reuse path against the real `create-worktree.sh` + `catalyst-thoughts.sh` + `worktree-thoughts-init.sh`):
- **8** — configured project + plain-dir `thoughts/shared` (the stranding bug) → repaired to a healthy symlink, stranded content preserved under `thoughts.orphaned-*`.
- **9** — configured project + *dangling* symlink → repaired (guards require `-L` AND `-d`, not just `-L`).
- **10** — project **without** shared thoughts → reuse still proceeds (exit 0), never hard-blocked; thoughts left untouched.

Full suite green locally (26/26, existing 1–7 unchanged).

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-thoughts-check.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-thoughts-check.test.sh
@@ -166,7 +166,13 @@ run_reuse_worktree() {
 		chmod +x "$BIN/humanlayer"
 		PATHX="$BIN:$PATH"
 	else
-		CONF='{"catalyst":{"projectKey":"t"}}' # no thoughts profile; humanlayer stays off PATH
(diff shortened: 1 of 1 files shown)


ryanrozich:
Fixed in the latest commit. Case 10 now shadows any ambient `humanlayer` with a stub that reports NO profile, so the reuse-path profile gate resolves empty and the repair is deterministically skipped — verified the case passes both with a clean PATH and with a profile-reporting `humanlayer` prepended (simulating CI). 26/26.

PR #2755, opened 2026-07-26
feat(dev): CTL-1534 phase-3 shadow consumer + smee/cloud parity harness

ryanrozich:
Phase 3 of the smee → Catalyst Cloud event-delivery cutover (ADR-0008 leg 3). CTL-1534.

**Deliberately a DRAFT.** The machinery is built, its negative controls are observed to fire, and the ten known defects listed in the original draft are now fixed — but the central claim is still not demonstrated. Details below. I'd rather this sit visible and honest than merge a harness that can report a false green.

## What's here

- **`cloud-event-consumer.mjs`** — polls the cloud `/events/stream`, dedups on `deliveryId`, maps each **raw** payload through the **unmodified** orch-monitor mappers (no second mapper — both paths must run identical mapping code, so any diff is transport or coverage, never mapping), stamps `webhook.delivery.id` + the feed seq, appends to a **shadow** log.
- **`parity-harness.mjs`** — joins shadow ↔ live on `webhook.delivery.id`; asserts **coverage and integrity separately**, per-source **non-zero** matched pairs, per-type **counts** (never presence) for the Gate B superset.
- Three-way exit on both (`0` healthy / `1` problem / `2` **could not evaluate**) and credential-free offline negative controls.

## Process

Built, then adversarially reviewed against the **11 real defects** catalyst-cloud found in its equivalent harness. The review found **35 (8 critical)** — every one *reproduced by execution*, not argued. All 8 criticals + 15 highs fixed and independently re-verified **by execution** (58 confirmed). A second round then closed the 10 defects that re-verification had left open (below).

The most instructive defect came from the build structure itself: the consumer never stamped a cloud-seq attribute, so the harness's entire seq subsystem keyed on an attribute that was never present — **evaluating nothing while passing**. Building the two halves in parallel produced exactly the false-green class the tool exists to prevent.

**The safety-critical fix, done by hand rather than delegated:** the shadow-dir guard was purely **lexical**, and `resolve()` does not follow symlinks. A symlinked `<root>/events-shadow → <root>/events` passed every check and would have written the **live** event log — double-waking every worker — while the safety self-test printed PASS. Now: refuses a symlinked shadow dir, re-checks on `realpath`, and writes `shadow-YYYY-MM.jsonl` so a shadow write can never even share the live basename. Regression tests reproduce the original attack.

## Round 2 — the ten known defects are FIXED

Each fix is pinned by a regression test that **fails on the pre-fix code**. Verified by mutation: every one of the eleven behaviours below was individually reverted and the suite went red.

| id | issue | fix |
|---|---|---|
| H | interior-page gap did **not** pin the cursor (only late-start did) → could advance past a proven hole | ✅ any proven hole pins the cursor and writes a durable `feed-gap` marker, wherever it sits in the page |
| H | marker `ts` was wall-clock while envelopes use `receivedAt` → markers fell outside their own window | ✅ markers anchor `ts` on `receivedAt`; the wall clock is kept separately as `emittedAt`, with `tsSource` naming which is which |
| M11 | a dedup-only pass (0 appended) still exited 0 | ✅ exit 2 — a pass that appended nothing confirmed nothing. Gated on an otherwise-healthy pass so a cursor pinned at a hole keeps its stronger, evaluated verdict |
| H10 | mapper-decline detector implemented only for *zero* appends; a **partial** regression was invisible | ✅ `classifyDecline` splits schema drift (structural) from a routine unhandled action; a structural decline on a declared-mappable type is a problem at **any** ratio |
| M | `provenHole` could assert a hole that doesn't exist under wire-order inversions | ✅ holes derive from the received **set** (`computeMissingRanges`), not `firstSeq`. Order and presence are separate questions; wire-order integrity is unchanged and still unsorted |
| M | durable marker writes sat outside the try/catch → wrong exit class on a write failure | ✅ single guarded `writeMarker`; a failed write is exit 2 with a named reason, never an escaped throw |
| M | edge-exclusion `not_run` was all-or-nothing; a partial exclusion still passed | ✅ `MISSING_FROM_SHADOW_EDGE_EXCLUDED` / `SHADOW_ONLY_EDGE_EXCLUDED` enumerate the waived candidates instead of folding them into a bare `pass` |
| M1 | config-read failures failed **open** (empty roster/bot set) instead of exit 2 | ✅ an empty roster or bot set is exit 2; a deliberate degradation must be **declared** via `--allow-empty-config`; a config file that exists but doesn't parse now throws |
| L | `seqOf()` coerced a string seq — evidence normalisation, the rule it violates | ✅ rejected as `SEQ_ATTR_INVALID`. The consumer stamps a number; a string means the producer changed shape |
| L | CI `bun test <file>` silently ignored a renamed/missing file and exited 0 | ✅ a guard step asserts every `bun test` argument resolves first. The list is parsed out of the workflow (never duplicated), resolution is by suffix (a bun-test arg is a filename filter), and a parse yielding nothing fails too |

**Plus the `KNOWN_GAP_LOSSES` alert-fatigue wrinkle:** it is now `info`, not `warn`. Every delivery it counts already matched a census row naming an open ticket, so `warn` held the harness permanently red for that ticket's whole life — which is precisely how a signal gets trained out of. The distinction the exit code must preserve is **new/uninvestigated** vs **known/filed**: an uncensused loss is not counted here at all and still **fails** `MISSING_FROM_SHADOW`. `--strict-known-gaps` escalates the tracked ones to a hard fail for the phase-5 close-out.

## Verification

- `bun test cloud-event-consumer.test.mjs parity-harness.test.mjs` → **272 pass, 0 fail** (was 243)
- `cloud-event-consumer.mjs --self-test` → **36 cases, 29 detectors expected red, 36 correct** (was 30/30)
- `parity-harness.mjs --self-test` → **34/34 detectors fired**, positive control green (was 31/31)
- Mutation check: each of the 11 fixed behaviours reverted individually → the suite goes red for each

## Still a draft — why

**The central claim is undemonstrated.** No non-zero shadow↔live join has been observed. The laptop's live log has no `webhook.delivery.id` rows (off-roster, monitor not running); `mini` has them only since CTL-1532. **This must be proven on `mini` before any parity claim** — a separate, deliberate step, not part of this round.

## Not done

- Not deployed anywhere.
- Gate A's ≥7-day parity window is calendar time, not work. This PR is the machinery plus evidence the machinery detects failure — **not** a completed parity run.

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -420,6 +420,14 @@ jobs:
         # exit-127'd the reaper fleet-wide. Runs the REAL refusal paths.
         working-directory: plugins/dev/scripts
         run: bash __tests__/install-orphan-sweep-guard.test.sh
+      - name: CTL-1534 negative controls (shadow consumer + parity harness self-tests)
+        # Rule 8: the load-bearing control must run where it is most useful — a clean,
+        # secret-free CI checkout. Both binaries exit 0 ONLY when every seeded detector
+        # fired and the positive control stayed green; a detector that stops firing fails
+        # the gate here instead of on a developer's laptop.
+        run: |
+          bun cloud-event-consumer.mjs --self-test
+          bun parity-harness.mjs --self-test
 
       - name: Run stable unit tests
         # EXCLUDED SUITES — do NOT add these back without fixing the underlying
@@ -486,6 +494,8 @@ jobs:
             tick-timing.test.mjs \
             tracing.test.mjs \
             cloud-token-env.test.mjs \
+            cloud-event-consumer.test.mjs \
+            parity-harness.test.mjs \
             cluster-claim-sync.test.mjs \
             cluster-claim.test.mjs \
             comms-drain.test.mjs \
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -429,6 +429,45 @@ jobs:
           bun cloud-event-consumer.mjs --self-test
           bun parity-harness.mjs --self-test
 
+      - name: Assert every `bun test` target below still exists (CTL-1534)
+        # `bun test <file>` on a path that does not exist matches ZERO tests and exits 0.
+        # So renaming or deleting any suite named in the allowlist below silently stops
+        # running it while this required check stays green — the exact "the check
+        # evaluated nothing and passed" shape the CTL-1534 tooling exists to prevent.
+        # The list is parsed out of this file rather than duplicated, so it cannot drift
+        # from the step it guards; a parse that yields nothing fails too (a guard that
+        # checks nothing is worse than no guard). Only the backslash-continued argument
+        # lines of the `bun test` invocation are read — the EXCLUDED-SUITES comment above
+        # it names suites that deliberately do NOT run.
+        #
+        # Resolution is by SUFFIX, not exact path: a `bun test` argument is a filename
+        # filter, so `config-node-class.test.mjs` legitimately matches
+        # `__tests__/config-node-class.test.mjs`. What must never happen is an argument
+        # that matches NOTHING.
+        working-directory: .
+        run: |
+          set -euo pipefail
+          targets="${RUNNER_TEMP:-/tmp}/bun-test-targets.txt"
+          awk '
+            index($0, "- name: Run stable unit tests") { f = 1; next }
+            f && /^      - name: / { f = 0 }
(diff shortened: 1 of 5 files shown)


ryanrozich:
Closing under the cloud-native pivot (ADR-0037 merged; ADR-0039 accepted 2026-08-23 — cloud is the single writer of coordination state; runners are cloud containers; resident-daemon mechanisms are being retired, and the event log moves to the DO-outbox→R2 backbone). Ryan pre-authorized draining PRs the new direction obsoletes. Reopen if the pre-cutover daemon genuinely needs this as a stopgap. — smee/cloud parity harness: smee retirement is explicit and the cutover decision already landed; stale 28d, author-flagged unproven.

PR #2798, opened 2026-07-30
fix(dev): CTL-1571 make queued/blocked labels retractable once a ticket starts

thagale:
## The problem

Two real pilot tickets sat for hours carrying a stale `queued` Linear label through multiple completed pipeline phases, long after being admitted and dispatched. The label served no purpose once the ticket started running, but nothing ever cleared it.

`convergeStartedHeldLabels` (CTL-1068) exists specifically to retract an orphaned held label from a STARTED ticket -- its own doc comment describes this exact scenario ("a ticket that was picked up and then failed wearing blocked/waiting is never revisited [by the admission loop], so its label lingers ... forever"). It is correctly wired into `schedulerTick` for every non-pre-pickup ticket, every tick.

But its retraction loop gates on the **existence of a local once-marker file** (`workers/<T>/.linear-label-<label>.{applied,skipped}`) before attempting anything -- and `queued`/`blocked` are never written through that marker convention. They are applied and removed entirely by `convergeHeldLabel`'s tick-diff in the pre-pickup admission loop, which uses a live label-set diff, not a marker file. So the one code path capable of finding and clearing a stuck `queued`/`blocked` label can never find one, because its precondition is never satisfied for these two labels -- even though the label really is present on Linear.

The failure is intermittent, not universal, which is why it slipped past most tickets: `convergeHeldLabel`'s clear-on-pickup write only gets ONE attempt, in the exact tick a ticket transitions from waiting to admitted. If that single `removeLabel` call fails for any transient reason, the ticket has already left the pre-pickup pool by the next tick and nothing ever retries it.

## The fix

`convergeHeldLabel` now writes the same once-marker `labelOnce` writes for `needs-human`, on confirmed apply, and deletes it on confirmed removal (mirroring #2796's "keep the once-marker in step with the label"). This gives `convergeStartedHeldLabels`' existing, already-tested, already-fenced retraction logic something to find. No new Linear reads, no new retry infrastructure -- just making the write side leave the breadcrumb the read side was always designed to look for.

## Verification

New test (CTL-1571) reproduces the bug end-to-end: applies `queued` via `convergeHeldLabel`, then calls `convergeStartedHeldLabels` as if the ticket had since started, and asserts the label is now retracted. Confirmed by mutation: reverting the `scheduler.mjs` fix makes this exact test fail at the marker-existence assertion (no source change reverts the test).

Full execution-core suite (`bun test`, 228 files): 7037 pass / 21 fail including this fix, vs 7039 pass / 18 fail on `main` -- the +3 delta isolates to +1 new passing test (this one) and a pre-existing flaky integration test (`execution-core integration — crash recovery (CTL-539) > AC5`) that did not reproduce on two further full-suite runs and passes cleanly in isolation regardless.

`scheduler.test.mjs` alone: 597 pass / 2 fail, unchanged from baseline's 596/2 plus this one new passing test.

Closes CTL-1571.

## Test plan
- [x] `bun test scheduler.test.mjs` -- 597 pass / 2 fail (2 pre-existing, unrelated: CTL-781 delegate tests)
- [x] `bun test` (full execution-core suite) -- no net-new regressions vs `main`, confirmed across 3 runs
- [x] Mutation-tested the new test against a reverted fix -- fails exactly at the expected assertion

### DIFF
--- a/plugins/dev/scripts/execution-core/scheduler.mjs
+++ b/plugins/dev/scripts/execution-core/scheduler.mjs
@@ -23,6 +23,7 @@ import {
   rmSync,
   renameSync,
   statSync,
+  unlinkSync,
 } from "node:fs";
 import { spawnSync } from "node:child_process";
 import { monitorEventLoopDelay } from "node:perf_hooks";
@@ -346,6 +347,7 @@ import {
   labelNeedsHumanUnlessBeliefOwner,
   resolveAndApplyWorkerStatusLabel,
   WORKER_STATUS_LABELS,
+  labelMarkerBase, // CTL-1571: convergeHeldLabel writes the same once-marker labelOnce does
 } from "./label-guard.mjs";
 import { DISPOSITIONS } from "./worker-disposition.mjs"; // CTL-1605: precedence order for the onTerminalCleared aggregate-arg → pre-clear `from` resolution
 import { processApprovedResumes } from "./boot-resume.mjs"; // CTL-644: per-tick approval poll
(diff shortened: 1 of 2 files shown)


ryanrozich:
Thanks for this, Tony — reviewed and it's merged via #3045 (had to migrate off the fork branch since Cloudflare Pages can't run required checks on fork heads — nothing about your patch, just repo plumbing). Really appreciate the contribution. We're working on getting you a CLA + direct contributor access so future PRs can land straight from a branch on the main repo.

PR #2818, opened 2026-07-31
fix(monitor): pin CATALYST_CONFIG_FILE in cmd_start so Layer-1 config never falls back to the operator's cwd

thagale:
## Summary

`cmd_start()` spawned the monitor server with only `CATALYST_CONFIG_PATH`
threaded through (defaulting to empty). Without `CATALYST_CONFIG_FILE` set,
`resolveLayer1ConfigPath()` falls back to `${cwd}/.catalyst/config.json` —
whatever directory the operator happened to be in when they ran
`catalyst-monitor start` / `catalyst-stack start`.

When that directory lacks `.catalyst/config.json`, the team/project roster
and the Layer-2 Linear-token resolution used by the reply endpoint both
silently degrade: an empty/wrong project list, board views that never
resolve, and replies failing with a missing-credential error — even though a
correctly-configured `$CATALYST_DIR/.catalyst/config.json` exists and would
have resolved fine.

Hit this live: a monitor process had been running for hours from the wrong
cwd, degrading the dashboard's project list and silently breaking the
Linear-reply credential path, with no error surfaced anywhere pointing at
the actual cause.

## Fix

Pin `CATALYST_CONFIG_FILE="${CATALYST_CONFIG_FILE:-$CATALYST_DIR/.catalyst/config.json}"`
in `cmd_start()`, only when the caller hasn't already pointed at a specific
file. `CATALYST_DIR` is already resolved earlier in the script
(`${CATALYST_DIR:-$HOME/catalyst}`), so this is a one-line change with no new
inputs.

## Test plan
- [x] Ran `catalyst-monitor-bootstrap.test.sh` (17 passed) and
      `catalyst-monitor-dist-redirect.test.sh` (21 passed) — including the
      existing `cmd_start injects MONITOR_PUBLIC_DIR into server env` test,
      unaffected by this change.
- [x] Verified live: restarted the monitor from the correct cwd, confirmed
      `/api/projects` returns the full configured team roster
      (`source:"config"`) instead of the degraded fallback.

### DIFF
--- a/plugins/dev/scripts/catalyst-monitor.sh
+++ b/plugins/dev/scripts/catalyst-monitor.sh
@@ -516,6 +516,17 @@ cmd_start() {
     linear_app_actor_auth "catalyst-monitor" CATALYST_MONITOR_APP_ACTOR_TOKEN
   fi
 
+  # CATALYST_CONFIG_FILE pins the Layer-1 config path explicitly so the spawned
+  # server's config resolution (orch-monitor/lib/config-path.ts) never falls back
+  # to a cwd-relative `.catalyst/config.json` lookup. Without this, the server
+  # inherits whatever directory the operator happened to be in when they ran
+  # `catalyst-monitor start` / `catalyst-stack start` — if that directory has no
+  # `.catalyst/config.json` (or the wrong one), the team/project roster and the
+  # Layer-2 Linear-token resolution both silently degrade (empty project list,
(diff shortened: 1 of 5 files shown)


ryanrozich:
Following up — this fix has already shipped to main via #3052 (CTL-1662/PR #2851's chain, which stacked on top of this commit too). Closing the original PR and its migration (#3056) as redundant. Thanks for the contribution and for surfacing this, Tony.

PR #2819, opened 2026-07-31
fix(execution-core): scope CTL-778's alive-but-idle reclaim to the current dispatch attempt

### LINKED ISSUE #2816: ctl-778 alive-but-idle reclaim kills redispatched phase workers via stale (unscoped) work-done evidence
## Summary

`ctl-778`'s "alive-but-idle" reclaim in `recovery.mjs` can kill a freshly-dispatched
phase worker within 5–10 seconds of launch, because the two signals it uses to
decide "this worker already finished" — `completeEventSeen()` and
`WORK_DONE_PROBES[phase]` — are keyed only on `(ticket, phase)`, with no scoping to
the *current* dispatch's generation/attempt/`bg_job_id`. If a phase previously
completed successfully and is later **redispatched** (an operator response to a
`needs-human` escalation, a retry, a manual resume, etc.), the prior run's on-disk
artifact and its `phase.<phase>.complete.<ticket>` event are still sitting in the
worker directory. Those look indistinguishable from genuine fresh completion, so
the brand-new worker gets reclaimed as "idle-but-done" almost immediately —
regardless of what it's actually doing.

## Where

- `recovery.mjs` — the alive-but-idle branch (~line 2494 in the current tree):
  ```js
  if (
    hasProbe(phase) &&
    completeEventSeen({ ticket, phase }) &&
    probes[phase]({ ticket, repoRoot, orchDir })
  ) {
    // ... reclaim: emitComplete + kill the live bg job
  }
  ```
- `event-scan.mjs` `hasCompleteEvent()` (~line 240): checks whether
  `phase.<phase>.complete.<ticket>` has *ever* appeared in the event log window —
  not whether it corresponds to the worker currently running.
- `work-done-probes.mjs` (e.g. `verifyProbe`, ~line 109; the same pattern applies to
  `triageProbe`/`reviewProbe`/`monitorDeployProbe`): validates artifact *shape*
  (`verify.json` has the right fields), but never checks the artifact's mtime
  against the current dispatch signal's `startedAt`.

## Repro

1. Let a phase (e.g. `verify`) complete successfully once — its worker directory
   now has a valid `verify.json` and a `phase.verify.complete.<ticket>` event on
   record.
2. Later, redispatch that same phase for the same ticket (revive after an
   escalation response, a retry after a stall, etc.) — a new `bg_job_id` starts,
   `phase-verify.json` flips to `status: "running"`.
3. Within one scheduler tick (observed ~5–10s), `daemon.log` logs:
   ```
   ctl-778: alive-but-idle worker reclaimed (complete event + probe)
   ```
   and the brand-new worker is killed before it produces any real output. If the
   old artifact happened to carry escalation content (e.g. a `stalled` /
   `needs-human` payload from the prior run), that stale content gets surfaced as
   if the new run produced it.

This repeats on every subsequent redispatch attempt until the stale artifact +
event are cleared out-of-band — the phase can never make forward progress on its
own.

## Suggested fix

Gate the alive-but-idle reclaim on freshness, not just presence: only treat
`completeEventSeen`/probe evidence as valid if the artifact's mtime (or the
matched event's timestamp) is `>=` the *current* signal's `startedAt`. A
genuinely-completing worker's artifact is always written during/after its own
run, so this should be a no-op for the healthy path, while closing the stale-
evidence race for redispatched phases.

An alternative (heavier) fix would be archiving/renaming the prior-generation
artifact when a phase claims a new generation in `phase-agent-dispatch`, so a
redispatch never sees stale terminal state at all — but that touches every
probed phase's dispatch path rather than the single reclaim decision point.

Happy to open a PR for the freshness-gate version if that's the fix shape
maintainers would want — wanted to confirm the read on root cause first.

thagale:
## Summary

Fixes #2816.

`completeEventSeen` (the disambiguator `reclaimDeadWorkIfPossible`'s
alive-but-idle branch uses between "done, reclaim" and "busy, leave alone")
only asked whether a `phase.<phase>.complete.<ticket>` event had **ever**
been observed, with no scoping to the specific dispatch attempt currently
running.

A phase that completed once and is later redispatched — an operator resume,
a retry, a revive — had its brand-new worker reclaimed within seconds of
spawning, because the prior attempt's complete event (and its matching
work-done artifact) still satisfied the check. It's indistinguishable from
the new worker genuinely finishing, so the phase can never make forward
progress on a redispatch — every subsequent attempt gets killed the same
way.

Confirmed live across multiple real recurrences on the same ticket,
including one where the stale evidence was only ~15 minutes old (from a
genuinely-completed prior attempt), not a long-stale artifact — ruling out
"just clear old files" as a durable fix.

## Fix

- Adds `latestCompleteEventTs` (`event-scan.mjs`) alongside the existing
  `hasCompleteEvent` — same underlying per-event index, just also returns
  the latest matching event's timestamp instead of a plain boolean.
- `completeEventSeen`'s default now accepts an optional `sinceIso`; when
  present, it compares against `latestCompleteEventTs` instead of
  `hasCompleteEvent`, so only a complete event **at or after** the current
  dispatch's own start counts as "this worker finished."
- The CTL-778 reclaim call site now passes `sinceIso: signal.raw?.startedAt`
  (the current signal's own start time).
- Callers that omit `sinceIso` keep the old unscoped "ever" behavior — this
  is opt-in, scoped to the one call site that needed it.

## Test plan
- [x] 6 new tests for `latestCompleteEventTs` (null on no/missing events,
      correct single-event ts, latest-wins across multiple events, ticket/
      phase isolation, missing-arg guards).
- [x] 2 new tests for the recovery.mjs wiring: `completeEventSeen` receives
      `sinceIso` = the current signal's `startedAt`; a stale complete event
      is correctly rejected while a fresh one is accepted.
- [x] Full `event-scan.test.mjs` + `recovery.test.mjs`: 362/362 passing, no
      regressions against the pre-existing CTL-778/CTL-736 test coverage.
- [x] Full execution-core suite (7027 tests): pre-existing failures only, all
      in files this change doesn't touch (missing `pino`/`codex` binaries,
      hostname-dependent tests, timing-flaky integration tests).

### DIFF
--- a/plugins/dev/scripts/catalyst-monitor.sh
+++ b/plugins/dev/scripts/catalyst-monitor.sh
@@ -516,6 +516,17 @@ cmd_start() {
     linear_app_actor_auth "catalyst-monitor" CATALYST_MONITOR_APP_ACTOR_TOKEN
   fi
 
+  # CATALYST_CONFIG_FILE pins the Layer-1 config path explicitly so the spawned
+  # server's config resolution (orch-monitor/lib/config-path.ts) never falls back
+  # to a cwd-relative `.catalyst/config.json` lookup. Without this, the server
+  # inherits whatever directory the operator happened to be in when they ran
+  # `catalyst-monitor start` / `catalyst-stack start` — if that directory has no
+  # `.catalyst/config.json` (or the wrong one), the team/project roster and the
+  # Layer-2 Linear-token resolution both silently degrade (empty project list,
+  # board views that never resolve, replies failing with "no Linear credential")
+  # even though a correctly-configured `$CATALYST_DIR/.catalyst/config.json`
+  # exists. Only set when the caller hasn't already pointed at a specific file.
+  CATALYST_CONFIG_FILE="${CATALYST_CONFIG_FILE:-$CATALYST_DIR/.catalyst/config.json}" \
   CATALYST_CONFIG_PATH="${CATALYST_CONFIG_PATH:-}" \
   MONITOR_PORT="$PORT" \
   MONITOR_PUBLIC_DIR="${MONITOR_UI_DIST_DIR}" \
--- a/plugins/dev/scripts/execution-core/event-scan.mjs
+++ b/plugins/dev/scripts/execution-core/event-scan.mjs
@@ -244,6 +244,32 @@ export function hasCompleteEvent({ ticket, phase, path = getEventLogPath() } = {
   return entry.completes?.has(name) ?? false;
 }
 
+// CTL-778 P2: the ISO `ts` of the MOST RECENT phase.<phase>.complete.<ticket>
+// envelope, or null if none exists. hasCompleteEvent only answers "ever" —
+// unscoped to any particular dispatch attempt, so a phase that completed once
+// and is later redispatched (revive, retry, operator resume) has its brand-new
+// worker misread as "already done" by the CTL-778 alive-but-idle reclaim, which
+// checks hasCompleteEvent with no attempt/generation scoping at all. Callers
+// that need "did THIS dispatch's worker complete" (not "did any worker, ever")
+// should compare this against the current signal's startedAt instead of calling
+// hasCompleteEvent directly. Reuses entry.events (already retains ts per event
+// for every complete/revive/remediate envelope — see refreshIndex) so this adds
+// no extra indexing cost. String comparison on `ts` is valid because every
+// envelope's timestamp is the same fixed-width ISO-8601 UTC shape (buildEventEnvelope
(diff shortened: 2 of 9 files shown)


ryanrozich:
Closing as superseded — your own consolidated fix stack in #2858 already contains this PR's commits plus the review-finding fixes. Thanks for driving that consolidation, Tony.

ryanrozich:
Closing as superseded — your own consolidated fix stack in #2858 already contains this PR's commits plus the review-finding fixes. Thanks for driving that consolidation, Tony.

PR #2821, opened 2026-07-31
fix(dev): pre-push safety gate for placeholder-identity + blast-radius commits

thagale:
## Summary

A phase-agent's Bash access + real push credentials means it can, in
principle, push something destructive with nothing catching it before it
reaches the remote. Catalyst already has staleness protection and
concurrent-write protection (force-with-lease) on pushes, but nothing
inspects the *content* of what's about to be pushed.

Confirmed live: a verify-phase agent investigating a bug covered by a
repo's own test suite (one that builds its own throwaway git sandbox —
`mkdtemp`, `git init`, placeholder identity, wipe-the-tree, commit as
`"fixture"` — to exercise git-touching behavior) re-ran that exact recipe
directly against its real working directory instead of an isolated sandbox.
It reached the real branch before an operator caught it. Confirmed not
malicious — the placeholder author identity is the opposite of what a real
attacker would use to blend in, and the timing traced cleanly to the
dispatch's own activity — but the near-miss exposed a real gap: nothing
would have stopped a genuinely bad push either.

## Fix

A pre-push safety gate in `draft-pr.sh` (`_draft_pr_safety_gate`, new
`rc=4`) that fail-closed refuses a push when the pending commit range
carries:
- a placeholder author/committer identity, or
- an anomalous whole-tree deletion with no offsetting insertions
  (blast-radius check).

Wired into both `draft_pr_push` and `draft_pr_push_verify`. `phase-pr`'s
`SKILL.md` translates `rc=4` into a `push_safety_gate_blocked` escalation at
both push-verify call sites, so a blocked push surfaces as an operator
escalation rather than a silent failure. Also adds a `phase-verify`
skill-doc constraint warning agents never to hand-run a test's own
git-fixture setup against their real cwd — the proximate mistake in the
incident above.

## Test plan
- [x] `draft-pr.test.sh` — new coverage for the safety gate (placeholder
      identity detection, blast-radius detection, rc=4 propagation).

### DIFF
--- a/plugins/dev/scripts/lib/__tests__/draft-pr.test.sh
+++ b/plugins/dev/scripts/lib/__tests__/draft-pr.test.sh
@@ -1294,6 +1294,86 @@ chmod +x "${P10B_BIN}/git"
 ) || true
 assert_contains "$(cat "$P10B_OUT" 2>/dev/null)" "password=ghp_benigntoken123" "10b: helper emits the token as the credential password"
 
+# ─── Suite 11: _draft_pr_safety_gate — placeholder identity + blast radius (postmortem) ─
+echo ""
+echo "Suite 11: _draft_pr_safety_gate"
+
+# 11a: a commit authored by a placeholder identity (e.g. a test's own throwaway
+#      git-fixture recipe run against the real tree) blocks the push. rc==4,
+#      no output, origin/feature never created.
+echo "11a: placeholder-identity commit → draft_pr_push_verify returns 4, no push"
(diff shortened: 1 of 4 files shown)


ryanrozich:
Closing as superseded — this commit is included verbatim in #2858, which also fixes the outstanding review findings. Thanks Tony.

ryanrozich:
Closing as superseded — this commit is included verbatim in #2858, which also fixes the outstanding review findings. Thanks Tony.

PR #2846, opened 2026-07-31
fix(execution-core): pin CATALYST_CONFIG_FILE in cmd_start so daemon config never falls back to the operator's cwd

thagale:
## Summary

`daemon.mjs`'s `main()` resolves its Layer-1 config path as:
```js
process.env.CATALYST_CONFIG_FILE || resolve(process.cwd(), ".catalyst", "config.json")
```
— the same class of bug as the monitor's `cmd_start` (see #2818). `cmd_start()`
never pinned `CATALYST_CONFIG_FILE` before spawning the daemon, so it
inherited whatever directory the operator happened to be in when they ran
`catalyst-execution-core start` / `catalyst-stack start`. When that
directory lacks `.catalyst/config.json`, every Layer-1-driven feature
(`executorByPhase` routing, the reaper/worktree-refresh/stale-PR-rescue
timers, reconcile config) silently reads empty defaults instead of the real
config — with nothing surfacing the mismatch anywhere.

Hit this live: added `executorByPhase` routing to a Layer-1 config,
restarted the daemon from the wrong cwd, and the routing was silently
invisible to the daemon (no warning, no error — it just wasn't there) until
restarted from the correct directory.

## Fix

Pin `CATALYST_CONFIG_FILE="${CATALYST_CONFIG_FILE:-$CATALYST_DIR/.catalyst/config.json}"`
in `cmd_start()`, only when the caller hasn't already pointed at a specific
file — identical shape to the monitor fix in #2818.

## Test plan
- [x] Ran `catalyst-execution-core.test.sh` — 30/30 passing, no regressions.
- [x] Verified live: restarted the daemon from the correct cwd, confirmed
      the `executorByPhase` boot-eligibility log line appeared (previously
      silent/absent).

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# Tests for catalyst_git_exclude_worktree_artifacts (create-worktree.sh):
+# Catalyst's own worktree-local runtime artifacts (thoughts/,
+# .catalyst/.workflow-context.json, etc.) must never show up as dirty/untracked
+# in a fresh worktree's `git status` — without ever touching the project's
+# TRACKED .gitignore. This is what makes a project's verify-phase rebase never
+# get refused by Catalyst's own noise (the root cause of a real stalled-ticket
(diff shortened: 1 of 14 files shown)


ryanrozich:
Closing as superseded — all commits here are included in #2858 along with the review-finding fixes. Thanks Tony.

ryanrozich:
Closing as superseded — all commits here are included in #2858 along with the review-finding fixes. Thanks Tony.

PR #2851, opened 2026-08-01
fix(dev): add missing 'triage' case to linear-transition.sh's default_state_for()

thagale:
## Summary

`default_state_for()` has a fallback target-state for every phase the daemon
transitions a ticket through (`research`/`planning`/`verifying`/`reviewing`/
`remediating`/`inReview`/`done`/...) except `triage`, which fell through to
the empty default and failed with `could not resolve target state
(transition='triage')`. Every other phase in the same table had this exact
kind of fallback; `triage` was a straightforward oversight.

Confirmed live across a fresh team: with `stateMap.triage` absent from the
project config (matches the shipped default `.catalyst/config.json`
template — no team has this key), every ticket's triage-start status write
failed this way. Non-fatal (best-effort, warn-and-continue — dispatch
proceeds regardless), so it's easy to miss unless something's actually
watching the daemon log for it.

## Fix

Add `triage) echo "Triage" ;;` to `default_state_for()`. This only fully
resolves once a team's own Linear Triage mode (`triageEnabled`) is turned
on — Catalyst's `applyTriageStatus` already defaults its expected-state
check to `"Triage"` (`registry.mjs`'s `resolveEligibleQuery` does the same),
so both sides already agreed on the target name; the local bash-side
fallback table was the missing piece connecting them.

## Test plan
- [x] New test (`linear-transition.test.sh` Test 33), mirroring the existing
      Test 22/23 pattern for `verifying`/`reviewing`: asserts the fallback
      resolves to `Triage` when `stateMap.triage` is absent from config.
- [x] Full suite: 65/65 passing (63 pre-existing + 2 new), no regressions.

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# Tests for catalyst_git_exclude_worktree_artifacts (create-worktree.sh):
+# Catalyst's own worktree-local runtime artifacts (thoughts/,
+# .catalyst/.workflow-context.json, etc.) must never show up as dirty/untracked
+# in a fresh worktree's `git status` — without ever touching the project's
+# TRACKED .gitignore. This is what makes a project's verify-phase rebase never
+# get refused by Catalyst's own noise (the root cause of a real stalled-ticket
+# incident: rebase_refused_dirty_tree on thoughts/ + .workflow-context.json).
+# Run: bash plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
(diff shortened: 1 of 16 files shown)


ryanrozich:
Thanks @thagale ! @codex please review

thagale:
Addressed the review findings above (1 P1, 4 P2) in #2858 — each verified against current code before fixing. See that PR's description for the finding-by-finding breakdown and test plan.

ryanrozich:
Closing as superseded — #2858 contains this PR's exact commits plus fixes for the P1/P2 findings Codex flagged here. Thanks for the fast follow-up, Tony.

ryanrozich:
Closing as superseded — #2858 contains this PR's exact commits plus fixes for the P1/P2 findings Codex flagged here. Thanks for the fast follow-up, Tony.

PR #2852, opened 2026-08-01
fix(dev): add worktree git-dir/common-dir to codex-exec writable_roots

thagale:
## Summary

`codex-exec`'s sandbox `writable_roots` allowlist never included a linked worktree's REAL git metadata directories, so `git commit` inside a `codex-exec`-routed `implement`/`remediate` phase was silently refused (or produced an empty commit / left work uncommitted) whenever the worktree was a linked worktree — which every ticket worktree is.

For a linked worktree (`git worktree add wt/<TICKET>`), the actual git state git needs to mutate on commit does **not** live inside the worktree tree:

- `HEAD`, `index`, `index.lock`, `COMMIT_EDITMSG` live under `--absolute-git-dir` (`<main-repo>/.git/worktrees/<name>/`)
- the object database + refs live under `--git-common-dir` (the main repo's shared `.git`)

Both are outside `worktreePath` and outside `orchDir`, so neither was ever added to `sandbox_workspace_write.writable_roots`, and `codex exec`'s sandbox refused writes to `index.lock` with a permission error — or, worse, `git commit` silently no-oped and the phase still reported `status: "done"` with the real work left uncommitted in the worktree.

## Fix

Add `resolveWorktreeGitDirs(worktreePath)` (mirrors the existing `resolveGitExcludePath` git-path-resolution style already in this file) and wire its `{ gitDir, commonDir }` into `resolveWritableRoots`. Best-effort — `spawnSync` failures/non-zero exits resolve to `null` and are filtered out by the existing `add()` dedup helper, so a missing/invalid worktree path degrades to the pre-fix behavior instead of throwing.

## Test plan

- New test in `codex-run-phase-agent.test.mjs`: builds a REAL linked worktree via `git worktree add` (not just `git init`, so `--absolute-git-dir` genuinely resolves outside the worktree tree — this is the actual bug scenario) and asserts `buildCodexArgs`'s resulting `writable_roots` contains both the real `--absolute-git-dir` and `--git-common-dir` paths.
- Confirmed the existing `buildCodexArgs` test (nonexistent `worktreePath`) is unaffected — `git rev-parse` against a nonexistent path fails, both resolve to `null`, filtered out, roots array unchanged.
- `bun test codex-run-phase-agent.test.mjs`: 55/55 passing.

### DIFF
--- a/plugins/dev/scripts/execution-core/codex-run-phase-agent.mjs
+++ b/plugins/dev/scripts/execution-core/codex-run-phase-agent.mjs
@@ -319,9 +319,39 @@ function resolveThoughtsRoot(worktreePath) {
   }
 }
 
+// resolveWorktreeGitDirs — the REAL git metadata paths for a linked worktree.
+// For a linked worktree (every ticket worktree under wt/<TICKET>), `git commit`
+// writes per-worktree state (HEAD, index, index.lock, COMMIT_EDITMSG) under
+// `--absolute-git-dir`, which lives OUTSIDE the worktree tree entirely (in the
+// main checkout's `.git/worktrees/<name>/`) — and updates the object database +
+// refs under `--git-common-dir` (the shared `.git`, also outside the worktree).
+// Neither was previously in the sandbox's writable_roots, so `git commit` inside
+// a codex-exec worktree was refused with a permission error on `index.lock`.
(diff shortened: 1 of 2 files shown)


ryanrozich:
Closing as superseded — this commit is included in #2858. Thanks Tony.

ryanrozich:
Closing as superseded — this commit is included in #2858. Thanks Tony.

PR #2853, opened 2026-08-01
fix(dev): git-exclude Catalyst's own worktree-local runtime artifacts

thagale:
## Summary

Every onboarded project had to hand-maintain its own tracked `.gitignore` entries for Catalyst's own worktree-local runtime bookkeeping: `thoughts/`, `.catalyst/.workflow-context.json(.bak)`, `.catalyst/worktree-provenance.json`, `.needs-cleanup`, `.orphaned_at`, `.trunk`. Miss one and `git status` shows Catalyst's own noise as dirty — which is exactly what stalled a real ticket's `verify` phase in production: the phase's pre-verify rebase refused to proceed against a tree that was only "dirty" because of `thoughts/` and `.workflow-context.json`, neither of which the project had ever been told to ignore.

Catalyst already has the right pattern for exactly this problem: `.agents/skills` (its own tooling symlink) is kept out of every project via the worktree's **local, uncommitted** `.git/info/exclude` (`codex-run-phase-agent.mjs`'s `gitExcludeAgents`/`resolveGitExcludePath`) — never the project's tracked `.gitignore`. That mechanism was (a) never extended to cover the other Catalyst-owned worktree-local artifacts, and (b) wired in only for the `codex-exec` executor path, not `bg`/`sdk`.

## Fix

Add `catalyst_git_exclude_worktree_artifacts` to `create-worktree.sh`, called unconditionally for every newly created worktree regardless of which executor will later run in it. It writes the same artifact-pattern list to the worktree's local git exclude (idempotent, best-effort — never aborts worktree creation on failure). No project ever needs a `.gitignore` PR for these again, retroactively or going forward.

## Test plan

- New test: `plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh` — builds a real scratch git repo + bare origin, runs `create-worktree.sh` for real, and asserts: (1) all seven patterns land in the fresh worktree's local git exclude, (2) seeding each artifact on disk produces zero `git status` noise, (3) calling the exclude function twice is idempotent (no duplicate lines). 12/12 assertions pass.
- Ran the full existing `create-worktree-*.test.sh` suite before and after this change (`git stash`) — identical failure counts (all pre-existing, environment-specific: this host's git `init.defaultBranch` defaults to `master`, unrelated to this change). Zero regressions introduced.

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# Tests for catalyst_git_exclude_worktree_artifacts (create-worktree.sh):
+# Catalyst's own worktree-local runtime artifacts (thoughts/,
+# .catalyst/.workflow-context.json, etc.) must never show up as dirty/untracked
+# in a fresh worktree's `git status` — without ever touching the project's
+# TRACKED .gitignore. This is what makes a project's verify-phase rebase never
+# get refused by Catalyst's own noise (the root cause of a real stalled-ticket
+# incident: rebase_refused_dirty_tree on thoughts/ + .workflow-context.json).
+# Run: bash plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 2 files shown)


ryanrozich:
Following up — the underlying feature (git-excluding Catalyst's own worktree-local runtime artifacts, with the worktree-scoped/root-anchored fixes) has now shipped to main via #3052 (CTL-1662/PR #2851's chain). Closing this original PR as well since #3050 (its migration) turned out to be fully redundant with what's already merged. Thanks Tony.

PR #2858, opened 2026-08-01
fix(dev): address upstream review findings on PR #2851

thagale:
## Summary

Addresses the 5 findings from the automated Codex review on #2851 (1 P1, 4 P2). Each was verified against the current code before fixing, not taken on faith from the review text.

## Findings addressed

1. **P1 — `create-pr/SKILL.md` push had no safety gate at all.** `phase-pr`'s "promote existing draft" path runs the placeholder-identity/blast-radius guard via `draft_pr_push_verify`; the "create a fresh PR" path (no existing draft) delegated straight to `create-pr`, whose own push step was a bare `git push` with zero guard. Wired `create-pr`'s push through the same `draft_pr_push_verify` helper every other push site in this plugin already uses.
2. **P2 — `linear-transition.sh`'s `triage` fallback hard-coded `"Triage"`.** A project with a customized `eligibleQuery.triageStatus` (e.g. `"Intake"`) has no reason to duplicate that into `stateMap.triage`, but the fallback ignored it — `applyTriageStatus()` verifies against the configured value, so this would either land in the wrong state or fail verification. Now resolves from the execution-core registry first.
3. **P2 — `catalyst-monitor.sh`'s `CATALYST_CONFIG_FILE` default broke the `CATALYST_CONFIG_PATH` back-compat contract.** Documented precedence (`config-path.ts`) is `CATALYST_CONFIG_FILE > CATALYST_CONFIG_PATH > cwd`; defaulting `CATALYST_CONFIG_FILE` whenever it was merely unset — regardless of whether the operator had explicitly set only the legacy `CATALYST_CONFIG_PATH` — silently overrode that. Now only defaults when neither var is already set.
4. **P2 — `catalyst-execution-core`'s same default broke the relocated-`CATALYST_DIR` case.** A deployment with `CATALYST_DIR` pointed at a state-only root (e.g. `/var/lib/catalyst`), started from the real repo's cwd, previously got a forced-nonexistent `CATALYST_CONFIG_FILE` instead of falling through to `daemon.mjs`'s own working cwd fallback — silent empty-defaults load, no error surfaced. Now only defaults when the `CATALYST_DIR`-relative candidate actually exists.
5. **P2 — CTL-778's stale-evidence guard had a same-second race.** `completeEventSeen`'s `ts >= sinceIso` compares second-precision timestamps, so a prior attempt's completion and a same-second redispatch's `startedAt` compare equal — reopening the exact bug CTL-778 was built to close. Added `latestCompleteEventAttempt` (event-scan.mjs) and thread the signal's own `attempt` through as `sinceAttempt` (recovery.mjs) as the precise discriminator. Fail-open: absent on either side, behavior is byte-identical to before.

## Test plan

- New/extended coverage for all five: `linear-transition.test.sh` (+2 tests), `event-scan.test.mjs` (+7 tests, including a same-second tie-break regression I caught myself while writing the new function — see commit), `recovery.test.mjs` (1 updated + 1 new).
- Full suites green for every touched file: `draft-pr` 85/85, `linear-transition` 67/67, `catalyst-monitor-bootstrap` 17/17, `catalyst-execution-core` 30/30, `event-scan` 59/59, `recovery` 311/311.

Addresses review on #2851.

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# Tests for catalyst_git_exclude_worktree_artifacts (create-worktree.sh):
+# Catalyst's own worktree-local runtime artifacts (thoughts/,
+# .catalyst/.workflow-context.json, etc.) must never show up as dirty/untracked
+# in a fresh worktree's `git status` — without ever touching the project's
+# TRACKED .gitignore. This is what makes a project's verify-phase rebase never
+# get refused by Catalyst's own noise (the root cause of a real stalled-ticket
+# incident: rebase_refused_dirty_tree on thoughts/ + .workflow-context.json).
+# Run: bash plugins/dev/scripts/__tests__/create-worktree-gitignore-artifacts.test.sh
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../../.." && pwd)"
+CREATE_WT="${REPO_ROOT}/plugins/dev/scripts/create-worktree.sh"
+
+FAILURES=0
+PASSES=0
+fail() {
+	FAILURES=$((FAILURES + 1))
+	echo "  FAIL: $1"
+}
+pass() {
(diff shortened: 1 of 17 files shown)


ryanrozich:
Thanks for this, Tony — reviewed and it's merged via #3052 (had to migrate off the fork branch since Cloudflare Pages can't run required checks on fork heads — nothing about your patch, just repo plumbing). Really appreciate the contribution. We're working on getting you a CLA + direct contributor access so future PRs can land straight from a branch on the main repo.

PR #2864, opened 2026-08-02
fix(execution-core): isTicketInFlight supersede guard for stale phase signals

### LINKED ISSUE #2866: isTicketInFlight vetoes a ticket forever on a stale predecessor phase, even after a later phase genuinely completes
## Problem

`readPhaseSignals` returns every `workers/<T>/phase-*.json` a ticket has EVER had, not just its current one. `isTicketInFlight` (`execution-core/scheduler.mjs`) vetoes a ticket the moment ANY of those historical signals is `failed`/`stalled`/`aborted` — with no notion of a later phase having genuinely superseded it.

This bites whenever a phase gets dispatched directly and out of order — e.g. an operator or agent re-dispatching a later phase to unstick a ticket (`phase-agent-dispatch --phase <later-phase> --ticket <T>`, the pattern the recovery-pass skill's Rubric Two explicitly documents for a "stalled phase died mid-flight" case), without also clearing the earlier phase's now-stale signal.

```gherkin
Scenario: A ticket's earlier phase failed transiently, but a later phase has since completed
  Given a ticket's workers/<T>/phase-implement.json is status:"failed" (a stale, superseded signal)
  And workers/<T>/phase-verify.json is status:"done" (real, later work that landed anyway)
  When the scheduler's advancement sweep runs
  Then isTicketInFlight should return true (the ticket has genuinely moved past implement)
  # CURRENTLY: it returns false, because implement's "failed" status is present in the
  # signal map and the check has no notion of it being superseded by verify's later "done"
```

**Consequence:** the ticket is silently walled off from `listInFlightTickets` and the entire advancement sweep — `verify`/`remediate` routing never gets (re-)checked, even once real, later work has genuinely landed. The ticket just sits there. The only visible symptom is the eyes+hands diagnostic sweep quoting the OLD failure reason from the superseded phase, which reads as a much older/different bug than the actual current state.

## Root cause

`isTicketInFlight` (`execution-core/scheduler.mjs`) iterates the full phase→status map and returns `false` on the first failed/stalled/aborted entry it finds, regardless of pipeline order:

```js
for (const [phase, status] of Object.entries(signals)) {
  if (status === "failed" || status === "stalled" || status === "aborted") return false;
  ...
}
```

There's no check for whether a *later* phase (by `phaseIndex`) has since been dispatched and succeeded.

## Fix

The codebase already has the exact primitive for this: the CTL-606 supersede guard (`recovery.mjs`'s reclaim path) already determines "is this phase signal superseded by a later-dispatched phase" using `phaseIndex`/`isKnownPhase` from `lib/phase-fsm.mjs`. `isTicketInFlight` should use the same check — a KNOWN phase whose `phaseIndex` trails the ticket's latest-dispatched KNOWN phase should not have its status counted toward the veto.

## PR

https://github.com/coalesce-labs/catalyst/pull/2864 (already merged into my fork: https://github.com/thagale/catalyst/pull/14) implements exactly this, reusing `phaseIndex`/`isKnownPhase`, with 8 new tests covering the supersede path (including `remediate`'s equal-index ranking with `verify`, and confirming unknown/non-pipeline signals like `recovery-pass` never participate in the ordering). All pre-existing `isTicketInFlight` tests pass unchanged; full `scheduler.test.mjs` suite verified via a stash/pop A-B comparison (507/101 before, 515/101 after — same 101 pre-existing unrelated failures both times, +8 new passes, 0 regressions).

Closes via #2864.

thagale:
Closes #2866

## Summary
`readPhaseSignals` returns every `workers/<T>/phase-*.json` a ticket has EVER had, not just its current one. `isTicketInFlight` vetoed a ticket the moment ANY of those historical signals was `failed`/`stalled`/`aborted`, with no notion of a later phase having genuinely superseded it.

This bites whenever a phase gets dispatched directly and out of order (e.g. an operator or agent re-dispatching a later phase to unstick a ticket, without also clearing the earlier phase's now-stale signal): the stale predecessor permanently vetoes `isTicketInFlight`, silently walling the ticket off from `listInFlightTickets` and the whole advancement sweep (verify/remediate never gets re-checked) even once real, later work has genuinely landed. The ticket just sits there — the only prior symptom was the eyes+hands diagnostic sweep quoting the OLD failure reason from the superseded phase, which reads as a much older/different bug than the real, current one.

Fixes it by reusing the existing CTL-606 supersede-guard primitive (`phaseIndex`/`isKnownPhase`, already used by `recovery.mjs`'s reclaim path) inside `isTicketInFlight` itself: a KNOWN phase whose `phaseIndex` trails the ticket's latest-dispatched KNOWN phase is superseded, and its status no longer counts toward the veto. Unifies the two "has this ticket moved past this signal" checks so they agree with each other.

## Test plan
- [x] 8 new tests cover the supersede path (incl. remediate's equal-index ranking with verify, and that unknown/non-pipeline signals like `recovery-pass` never participate in the ordering)
- [x] All pre-existing `isTicketInFlight` tests pass unchanged
- [x] Full `scheduler.test.mjs`: 515 pass / 101 fail before and after (same 101 pre-existing, unrelated failures on both a clean checkout and with this change, verified via stash/pop A-B comparison) — this change adds exactly 8 passes and 0 new failures

Already merged into my fork ([thagale/catalyst#14](https://github.com/thagale/catalyst/pull/14)) and running there.


thagale:
Closing as superseded — this exact commit was migrated to a same-repo branch and merged as #3081. No further action needed here.

PR #2896, opened 2026-08-03
fix(dev): orch-monitor — mobile nav trigger + Linear reply-token identity walk

thagale:
## Summary
Two independent orch-monitor fixes, already merged into thagale/catalyst (#18, #19), now proposed upstream:

**1. Restore sidebar nav trigger for narrow viewports.** CTL-1003 removed the header's `SidebarTrigger` for desktop chrome cleanup (keyboard-only: `[` / Cmd-B). Below the `md` breakpoint the sidebar renders as an off-canvas Sheet with no keyboard available (iOS Safari) and no other tap target — the `SidebarRail` drag-handle is itself desktop-only (`sm:flex`) — leaving mobile users with no way to open navigation at all. Restores the trigger, gated to `md:hidden` so desktop's keyboard-only chrome is unchanged.

**2. Try every configured Linear token's identity, not just the first.** The Inbox reply/unblock feature (`lib/linear-comment.mjs`, CTL-1569) refuses to post when its Linear token resolves to the app actor (correct — CTL-1567 ignores app-authored comments). But it only ever resolved and identity-checked ONE token (first non-empty string across env → Layer-2 config), and `LINEAR_API_TOKEN`/`LINEAR_API_KEY` aren't an exclusively-personal slot: `lib/linear-app-actor.sh` exports the app-actor's own OAuth token into those same env vars for any daemon needing bot credentials. On a host where the monitor also sources that script, its env always carries a non-empty *bot* token, permanently shadowing a correctly-configured personal `linear.apiToken` in Layer-2 config. `linearTokenCandidates()` now returns every configured candidate, deduplicated; `postOperatorComment` identity-checks each in priority order and posts with the first genuine human, skipping (not failing on) any bot.

## Test plan
- [x] `bun run typecheck` — clean
- [x] `bun test` (app-shell.test.tsx, __tests__/app-shell.test.ts, linear-comment.test.mjs) — 55 pass, 0 fail
- [x] Both commits already shipped and running clean on thagale/catalyst since merge (thagale/catalyst#18, #19)

Claude-Session: https://claude.ai/code/session_0126MqxP29TnmkskRPGXgzsQ

### DIFF
--- a/plugins/dev/scripts/orch-monitor/__tests__/app-shell.test.ts
+++ b/plugins/dev/scripts/orch-monitor/__tests__/app-shell.test.ts
@@ -214,9 +214,13 @@ describe("one header: no search box, no SidebarTrigger collapse icon (CTL-1003)"
     expect(shellSrc).toContain("CommandDialog");
   });
 
-  it("the SidebarTrigger collapse icon + its Separator are removed from the header", () => {
-    // `[` / Cmd-B still toggle via shouldToggleSidebar — only the visible icon goes.
-    expect(shellSrc).not.toContain("SidebarTrigger");
+  it("the SidebarTrigger collapse icon's Separator stays removed; the icon itself is DESKTOP-hidden, not gone (mobile-nav fix)", () => {
+    // `[` / Cmd-B still toggle via shouldToggleSidebar on desktop. CTL-1003 removed
+    // the always-visible SidebarTrigger; the mobile-nav follow-up restored it, but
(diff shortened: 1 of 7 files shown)


ryanrozich:
Thanks for this, Tony — reviewed and it's merged via #3054 (had to migrate off the fork branch since Cloudflare Pages can't run required checks on fork heads — nothing about your patch, just repo plumbing). Really appreciate the contribution. We're working on getting you a CLA + direct contributor access so future PRs can land straight from a branch on the main repo.

PR #2923, opened 2026-08-03
fix(execution-core): codex-exec thoughts symlink resolution + no-progress escalation count

thagale:
## Summary
Three related fixes in execution-core's dispatch/escalation path, found while investigating a batch of failed tickets in a real multi-tenant deployment. Already merged to thagale/catalyst#20 + thagale/catalyst#21; proposing upstream per the fork-and-upstream default.

**1. resolveThoughtsRoot never actually resolved the thoughts symlink target.** The real on-disk convention (`lib/provision-thoughts.sh`) is: `thoughts/` is a real directory whose immediate entries (`global`, `shared`) are themselves symlinks pointing outside the worktree — not "`thoughts` itself is a symlink" as the function assumed. Since `thoughts` isn't a symlink in the real shape, `realpathSync` just returned the unchanged worktree-local path, so the actual external write target never made it into codex's sandbox `writable_roots`. This has existed since codex-exec's original build (CTL-1457, 2026-07-14) with zero test coverage, only surfacing once research/plan phases actually started routing through codex-exec — which broke every research-phase artifact write for tickets caught in that routing.

**2. no-progress escalations always reported "0 attempt(s)" regardless of real retry history.** `escalateOnce`'s `finalAttemptCount` parameter is used both for the human-facing message AND (separately) the `attempts` array, which is built from the real ask history. For "no-progress" specifically, the caller passes a progress *quantity* (commits-ahead / doc size) instead of an attempt count — since the escalate decision fires when that quantity is unchanged, it's almost always 0. Confirmed across multiple real tickets: identical contradiction every time (3 real timestamps in `attempts`, "0 attempt(s)" in the message). Diagnosed as cosmetic-only — the actual 3-strikes retry/ask-cap logic runs and enforces correctly before this message is built — but still misleading for anyone trying to understand an escalation before answering it.

**3. no-progress escalations couldn't distinguish "genuinely stuck" from "the Claude account backing this bg executor hit its usage/session limit."** Root-caused a batch of 11 escalation-ask-cap tickets across a multi-host fleet to a single cause: the account hit its rate limit, and every retry attempt just replayed the harness's own "You've hit your session limit" reply with zero real work — which the existing progress gate correctly saw as "no forward progress" but had no way to label accurately. Added `detectSessionRateLimitHit`, which inspects the dead worker's own session transcript (tail-only, bounded) for that literal harness reply and, when found, gives the escalation an accurate `call_to_action` ("wait for the usage window to reset, or reroute to a different executor") and `observed.likely_cause: "account-rate-limited"` instead of the generic no-progress message.

Adding that new transcript read also tripped `event-log-read-guard.test.mjs`'s (CTL-1529) module-global taint-scan into flagging two pre-existing, unrelated reads (`readExecCoreBootEpoch`'s `daemon-boot.json`, `readBootEpoch`'s `/proc/stat`) that had never been flagged before — a false-positive side effect of the detector's over-approximation (its own header prefers this tradeoff). Allowlisted both with an honest reason + marker comments, matching the guard's own sanctioned remedy.

## Test plan
- [x] `codex-run-phase-agent.test.mjs` — 5 new tests (real on-disk shape with nested symlinks, legacy direct-symlink shape preserved, no-symlink dir, no thoughts dir, dangling symlink)
- [x] `recovery.test.mjs` — 1 regression test proving a 3rd-cycle no-progress escalation reports `final_attempt_count: 3` (matching `attempts.length`), not the old `0`/progress-quantity value; 15 new tests for `detectSessionRateLimitHit` (unit + wiring)
- [x] `event-log-read-guard.test.mjs` — 40/40 pass with the new allowlist entry
- [x] All three files: 427 pass, 0 fail

**Note on scope**: this PR is deliberately narrower than what shipped on the fork — it excludes the git-dir/common-dir writable-roots addition (already tracked separately as an open upstream PR, #2852) since `main` doesn't have that feature yet; this PR's `resolveWritableRoots` only adds the thoughts-symlink resolution.

Claude-Session: https://claude.ai/code/session_0126MqxP29TnmkskRPGXgzsQ

### DIFF
--- a/plugins/dev/scripts/execution-core/codex-run-phase-agent.mjs
+++ b/plugins/dev/scripts/execution-core/codex-run-phase-agent.mjs
@@ -306,22 +306,74 @@ function resolveDevPluginRoot(pluginDirs) {
   return pluginDirs.find((p) => typeof p === "string" && p.length > 0);
 }
 
-// resolveThoughtsRoot — the REAL path the worktree's `thoughts/` symlink points
-// to (it points OUTSIDE the workspace — see the protocol doc). Added to the
-// writable roots so codex can write research/plan artifacts under thoughts/.
-// null when there is no thoughts symlink (best-effort — never throws).
-function resolveThoughtsRoot(worktreePath) {
-  if (!worktreePath) return null;
+// resolveThoughtsRoots — the REAL path(s) any symlink under the worktree's
+// `thoughts/` points to (they point OUTSIDE the workspace — see the protocol
+// doc). Added to the writable roots so codex can write research/plan artifacts
+// under thoughts/. Best-effort throughout: a missing thoughts/ dir, an
+// unreadable directory, or a broken/dangling symlink is skipped rather than
+// thrown — never lets a sandbox-roots computation crash a dispatch.
+//
+// 2026-08-03 fix: the ACTUAL on-disk convention (lib/provision-thoughts.sh) is
+// NOT "thoughts is a symlink" — it's "thoughts is a REAL directory whose
+// immediate entries (`global`, `shared`) are themselves symlinks pointing
+// outside the worktree." The prior version only ever `realpathSync`'d the
+// `thoughts` directory itself; since `thoughts` isn't a symlink in that shape,
+// that just returned the unchanged worktree-local path and never added the
+// actual external target at all. Confirmed live: this silently broke every
+// research/plan artifact write for codex-exec-routed phases (4+ real tickets,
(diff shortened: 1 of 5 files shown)


ryanrozich:
Thank you for this contribution! This has been superseded by and merged via #3082 (which carried the fix forward through the fork→same-repo migration, plus additional hardening from a round-2/3 Codex security review on the thoughts-symlink writable-roots resolution). Closing this PR — the work is landed. Much appreciated!

PR #2935, opened 2026-08-03
fix(execution-core): close test-hermeticity gaps that only fail on a live, configured host

thagale:
## Summary
Investigating a batch of execution-core test failures that only reproduce on a real, already-configured multi-host Catalyst install — never on a clean CI runner, which is exactly why they went unnoticed. Already merged to thagale/catalyst#22; proposing upstream per the fork-and-upstream default.

Root-caused to test-hermeticity gaps, all in test-only files:

- `test-setup.mjs` (the `[test].preload`) pins `CATALYST_DIR` for hermeticity but never pinned `CATALYST_LAYER2_CONFIG_FILE` the same way. On a live host, unset falls through to the REAL `~/.config/catalyst/config.json` — including a genuine multi-host `catalyst.cluster.staticRoster` on any actual fleet node — silently turning "single-host" test setups into "multi-host," which then applies real HRW ownership filtering inside `schedulerTick`'s `ready.filter()`. This is what suppressed `integration.test.mjs`'s AC5 dispatch assertion on such a host.
- `config.test.mjs`'s "degrades to single-host" cases `delete`d the env var instead of pointing it at a guaranteed-absent path, leaking the same real roster into `getClusterHosts()` assertions.
- Both `config.test.mjs` and `heartbeat-event.test.mjs` asserted `hostname().replace(/\.local$/, "")` against `getHostName()`'s actual behavior (strip everything after the *first* dot) — the two formulas only agree on a bare label or a `*.local` host, and diverge on any real multi-label FQDN.
- `heartbeat-event.test.mjs`'s `readClusterHeartbeats` tests never passed `roster: []`, so on a real multi-host machine they picked up a genuine cross-host peer read merged into what should have been a local-only assertion.

All fixes make the failing assertions machine-independent rather than special-casing any particular host. No production code paths are touched — every changed line is inside a `.test.mjs` file or the shared test preload.

## Test plan
- [x] CI's exact curated test list (`execution-core-tests.yml`'s `bun test` invocation, upstream's current 153-file list): **5304/5304 pass, 0 fail**
- [x] Verified each fix's root cause with a standalone repro before patching (traced the AC5-style dispatch suppression to `schedulerTick`'s `multiHost` roster resolution via a scratch debug script, deleted after use)

Claude-Session: https://claude.ai/code/session_018Hj1J7kfaQrPNh5RVzAPa9

### DIFF
--- a/plugins/dev/scripts/execution-core/config.test.mjs
+++ b/plugins/dev/scripts/execution-core/config.test.mjs
@@ -10,6 +10,16 @@ import { describe, test, expect, beforeEach, afterEach } from "bun:test";
 import { mkdtempSync, writeFileSync, rmSync, mkdirSync, existsSync } from "node:fs";
 import { tmpdir, hostname, homedir } from "node:os";
 import { join, resolve } from "node:path";
+
+// Mirrors getHostName()'s fallback: strip everything after the FIRST dot, not
+// just a trailing ".local". A naive `.replace(/\.local$/, "")` only agrees with
+// this on a bare-label or *.local hostname — it diverges on any real multi-label
+// FQDN that isn't ".local" (e.g. this machine's "aldebaran.hagale.net"), which is
+// exactly the shape a live, non-mDNS Catalyst fleet host has.
+function firstDnsLabel(raw) {
+  const dot = raw.indexOf(".");
+  return dot === -1 ? raw : raw.slice(0, dot);
+}
 import {
   readWaitWatcherConfig,
   EVENT_DEBOUNCE_MS,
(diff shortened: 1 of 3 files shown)


ryanrozich:
Thanks for this, Tony — reviewed and it's merged via #3047 (had to migrate off the fork branch since Cloudflare Pages can't run required checks on fork heads — nothing about your patch, just repo plumbing). Really appreciate the contribution. We're working on getting you a CLA + direct contributor access so future PRs can land straight from a branch on the main repo.

PR #2987, opened 2026-08-04
feat(dev): dependabot-escalate — file a ticket for the two Dependabot signals no PR ever surfaces

thagale:
New sweep script + LaunchAgent installer. Auto-merge already covers minor/patch bumps fleet-wide (found during a 2026-08-04 dependabot-health audit); this adds the missing hard-case escalation path: a failed `Dependabot Updates` run (real dependency conflict, no PR to react to) or a PR labeled `major-update` now files a Linear ticket in the affected repo's team, idempotently.

Repo/team mapping is a local, uncommitted config file (not baked into the public script) — see `dependabot-escalate-repos.example.json` for the expected shape.

Verified stable across 3 consecutive real runs (0 new tickets after the first).

Also fixes two bugs found along the way, both in the local (untracked, `.git/hooks/pre-push`) safety net that scans pushes for private repo/org/ticket references:
- The range-computation bug that blocked PR #23 (fixed there, carried forward here).
- A stale denylist missing several fleet repos onboarded since it was first written, plus one overreaching bare term (`dit`, which matched ordinary English words like "edit"/"audit"/"credit").

PR #2988, opened 2026-08-04
fix(dev): drop pointless async from a synchronous server.stop() lifecycle test

thagale:
Unblocks the `quality` lint gate on other PRs that happen to touch `orch-monitor/` (found while shipping #2987 — the plist file it adds lives under `orch-monitor/dist/`, which incidentally triggers this path-filtered gate and surfaced this pre-existing error unrelated to that PR). The test never awaits anything and explicitly tests synchronous behavior (see its own comment); `async` was vestigial. Verified test still passes and lint goes from 1 error/92 warnings to 0 errors/92 warnings.


ryanrozich:
Thanks for this, Tony — checked and the exact same one-line change (dropping the pointless `async` from the `server.stop()` lifecycle test at server.test.ts:2197) already landed on `main` independently via #3011, so this PR's diff is now a no-op against `main` and there's nothing left to merge. The end state you were after is in place. Really appreciate the contribution. We're working on getting you a CLA + direct contributor access so future PRs can land straight from a branch on the main repo.

PR #3003, opened 2026-08-05
fix(dev): CTL-1648 — surface attentionReason in board stalled-phase derivation

ryanrozich:
## Summary

The orch-monitor board was displaying "gave up — unknown reason" for stalled tickets even when their signal files carried a specific `attentionReason` (e.g., `sdk-overloaded-exhausted`, `preempted-by-priority`). This fix aligns the board derivation with the scheduler's existing precedence — `failureReason ?? attentionReason ?? stalledReason ?? null` — across all four derivation sites in `board-data.mjs`, so the real reason is surfaced to operators instead of the opaque fallback.

### `plugins/dev/scripts/orch-monitor/lib/board-data.mjs` (+6 / -4)

Four derivation sites in `deriveCurrentPhase` and `derivePhaseWithRemediate` updated to use:

```js
failureReason: sig.failureReason ?? sig.attentionReason ?? sig.stalledReason ?? null
```

Previously, `attentionReason` was skipped entirely. The new order matches `scheduler.mjs:2666`'s `readDispatchFailureReason` — the same field the SDK launch backstop and `mark_launch_failed` write deliberately (a `failureReason` would trip the revive loop's non-retrying escalate branch, so `attentionReason` is the correct writer choice). This is a read-side-only change; no signal writers are touched.

### `plugins/dev/scripts/orch-monitor/lib/board-data.d.mts` (+1)

Added `failureReason?: string | null` to `BoardCurrentPhase` — the field was being set by the derivation but wasn't declared in the type.

### `plugins/dev/scripts/orch-monitor/board-data-failed-phase.test.ts` (+52)

Five new cases covering `deriveCurrentPhase` attentionReason surfacing:
- Non-terminal stalled phase with `attentionReason` only
- Terminal failed phase with `attentionReason` only
- Precedence: `failureReason` still wins over `attentionReason`
- Legacy: `stalledReason` still surfaces when it's the only field
- No fields set → `failureReason` is `null` (unknown-reason fallback intact)

### `plugins/dev/scripts/orch-monitor/__tests__/board-remediate.test.ts` (+30)

Three new cases covering `derivePhaseWithRemediate` attentionReason surfacing:
- Case 1 (remediate running): surfaces `remediate.attentionReason`
- Case 2 (remediate terminal, wins by updatedAt): surfaces `remediate.attentionReason`
- Precedence: `remediate.failureReason` still wins over `attentionReason`

## How to Verify

- [x] `tsc --noEmit` passes — type-clean ✅
- [x] `bun test board-data` passes — 330 tests pass (48 in modified files, 8 new CTL-1648 cases) ✅
- [x] `eslint board-data.mjs` passes — 0 errors ✅
- [ ] Manual: open a ticket in the stalled bucket whose signal has `attentionReason` but no `failureReason` — board should show the reason string instead of "unknown reason"

## Related Issues/PRs

- Fixes https://linear.app/coalesce-labs/issue/CTL-1648

### DIFF
--- a/plugins/dev/scripts/orch-monitor/__tests__/board-remediate.test.ts
+++ b/plugins/dev/scripts/orch-monitor/__tests__/board-remediate.test.ts
@@ -243,3 +243,33 @@ describe("ticketColumns — remediate ticket routes to remediate column", () => 
 test("REMEDIATE_PHASE export equals 'remediate'", () => {
   expect(REMEDIATE_PHASE).toBe("remediate");
 });
+
+// CTL-1648: remediate derivation must surface attentionReason when failureReason is absent
+describe("derivePhaseWithRemediate — attentionReason surfacing (CTL-1648)", () => {
+  it("Case 1 (remediate RUNNING): surfaces remediate.attentionReason", () => {
+    const s = sigs();
+    s[idx("verify")] = { status: "done", updatedAt: "2026-08-05T01:00:00Z" };
+    const remediateSig = { status: "running", attentionReason: "ctl-587-revive-reset",
+      startedAt: "2026-08-05T01:05:00Z", updatedAt: "2026-08-05T01:06:00Z" };
+    expect(derivePhaseWithRemediate(s, remediateSig).failureReason)
+      .toBe("ctl-587-revive-reset");
+  });
+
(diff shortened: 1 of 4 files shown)


ryanrozich:
@codex review

PR #3072, opened 2026-08-07
fix(ci): let execution-core-unit-tests always resolve when required

thagale:
## Summary

A required status check that's also path-filtered at the trigger level can never
resolve on a PR whose diff doesn't match the filter — GitHub has no conclusion to
post, so branch protection waits on "Expected" forever regardless of what the PR
touches. Hit this on a downstream fork PR whose only change was a config file
outside every path in this workflow's filter: `mergeable: MERGEABLE` per GitHub,
but `mergeStateStatus: BLOCKED` on a check that had never run and never would.

- `pull_request` now triggers unconditionally.
- A new `detect-changes` job (`dorny/paths-filter`, same glob set the `push`
  trigger already enforces) computes whether execution-core is actually relevant.
- `execution-core-unit-tests` gates on that via a job-level `if:` instead of the
  trigger-level `paths:`. A job skipped via `if:` still reports a real `skipped`
  conclusion, which GitHub does accept for a required check — unlike a workflow
  that never triggered at all.
- Kept as a job-level condition rather than removing path-scoping entirely: the
  suite is substantial (~30 steps — 5 import-graph builds, ~15 bash suites, one
  aggregate ~150-file `bun test` run), real cost to pay on every PR regardless of
  relevance.
- `push` keeps its existing path-filtered trigger unchanged (informational only,
  doesn't gate merges).

Proposing here per the dual-PR policy — this same workflow file is identical
between here and the fork today, so the same deadlock would eventually hit a PR
here too.

## Test plan

- [x] `actionlint .github/workflows/execution-core-tests.yml` — clean
- [x] `python3 -c "import yaml; yaml.safe_load(open(...))"` — parses, structure verified
- [x] Verified on the downstream fork: after this fix, the previously-BLOCKED PR's
      required check resolved for the first time (see thagale/catalyst#30)

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -1,7 +1,11 @@
 # execution-core-tests.yml
 #
 # Runs the stable execution-core unit-test suite on every PR and push to main
-# that touches the execution-core source tree.
+# that touches the execution-core source tree. The workflow itself now always
+# triggers on `pull_request` (see the PR-TRIGGER NOTE below); `detect-changes` +
+# `execution-core-unit-tests`'s job-level `if:` is what keeps the actual suite
+# scoped to relevant changes — a PR that doesn't touch execution-core gets a fast
+# `skipped` conclusion instead of paying for the full ~30-step job.
 #
 # REQUIRED-STATUS-CHECK OPERATOR NOTE (CTL-963)
 # -----------------------------------------------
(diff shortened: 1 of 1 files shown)


ryanrozich:
Superseded by #3107 — this PR was opened from a fork (\`thagale/catalyst\`), and fork PRs against this repo can never pass the required Cloudflare Pages check (it doesn't build fork branches), so this PR stays permanently \`BLOCKED\` regardless of review state. #3107 pushes the identical head content to a base-repo branch (so Cloudflare Pages can actually run) and additionally fixes the outstanding Codex P1 finding (fail-closed change detection). Closing this one in favor of #3107.

PR #3075, opened 2026-08-07
fix(dev): make match_thoughts_artifact shell-agnostic (zsh/shopt bug)

thagale:
## Summary

Same fix as thagale/catalyst#33, proposed here per the dual-PR policy since this code is shared.

\`shopt -s nullglob nocaseglob\` is a bash builtin. This file has a bash shebang
but in practice gets sourced under an interactive zsh session rather than
invoked via \`bash -c\` as intended. Under zsh, \`shopt\` silently fails
(non-fatal), and the subsequent bash array-glob assignment then hits zsh's own
default no-match behavior (errors rather than expanding to nothing) on
whichever alternative pattern doesn't match — poisoning the WHOLE assignment,
so even an alternative that DID match never made it into the array.

Rewrote \`match_thoughts_artifact\` with \`find -iname\`, which performs the
identical slug-tolerant, boundary-safe, case-insensitive match without
depending on either shell's array/glob-option extensions.

## Test plan

- [x] Reproduced the bug live under zsh, confirmed the fix resolves it
- [x] \`bash plugins/dev/scripts/__tests__/phase-artifact-gate.test.sh\` — 26/26 pass
- [x] \`bash plugins/dev/scripts/lib/__tests__/phase-artifact-gate.test.sh\` — 13/13 pass
- [x] \`bash plugins/dev/scripts/__tests__/phase-skill-globs.test.sh\` — 5/5 pass
- [x] \`shellcheck\` — clean

### DIFF
--- a/plugins/dev/scripts/__tests__/phase-artifact-gate.test.sh
+++ b/plugins/dev/scripts/__tests__/phase-artifact-gate.test.sh
@@ -153,6 +153,28 @@ assert_eq "0" "$RC10" "slugged plan doc satisfies the gate (writer↔gate agreem
 assert_contains "$OUT10" "phase-artifact-gate-contracts.md" "slugged plan doc found by matcher"
 
 echo ""
+echo "Test 11 (regression): match_thoughts_artifact behaves identically sourced under zsh"
+# `shopt` is not a zsh builtin, and the prior bash-array-glob implementation
+# relied on it (`shopt -s nullglob nocaseglob`). Under zsh that call silently
+# failed (non-fatal), then the array-glob assignment hit zsh's own default
+# no-match behavior (error, not empty-expand) on the FIRST alternative pattern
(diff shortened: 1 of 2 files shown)


ryanrozich:
Closing — this PR was opened from a fork (\`thagale/catalyst\`), which is permanently BLOCKED for this repo: the main branch ruleset requires the Cloudflare Pages check, and Cloudflare Pages does not build fork branches, so \`mergeStateStatus\` stays BLOCKED forever regardless of review/CI state (see #2905 for the same failure mode).

Migrated the exact head commit to a base-repo branch and fixed the outstanding Codex round-1 findings (portable \`find\` depth-limiting via \`-prune\` instead of \`-maxdepth\`, deterministic sort order, the committed \`CAT-39\` identifier, and CI wiring for the zsh regression test) on the superseding PR: #3108.

PR #3129, opened 2026-08-08
feat(dev): PR review convergence policy in phase-review and review-comments

thagale:
## Summary
- Every review round fixes P1 findings inline.
- From round 2 onward, P2-and-lower findings are filed to a follow-up ticket (verbatim, one per PR) instead of being chased indefinitely — `phase-review` previously recorded MEDIUM/LOW findings with no follow-through, and `review-comments` had no round awareness at all.
- `phase-review`: MEDIUM/LOW findings now file/append a follow-up ticket, recorded on `review.json`'s new `followUpTicket` field.
- `review-comments`: tracks review round via prior "address review comments" commits; round 1 stays unrestricted, round 2+ defers P2-or-lower findings to a new Step 3b (ticket + thread reply + resolve) instead of Step 3 (inline fix).

## Test plan
- [x] Reviewed both SKILL.md diffs for internal consistency (severity vocabulary, findings shape, cross-references between the two skills)
- [ ] Live-tested against a real multi-round PR


thagale:
Closing — per Tony, we're not pushing this to catalyst upstream right now. Landed via thagale/catalyst#34 (fork) instead.

PR #3146, opened 2026-08-08
feat(dev): CTL-1679 — recovery pass retries safe/retryable failures by default

ryanrozich:
## Summary

The recovery pass now **retries safe, retryable failures by default** instead of silently skipping
unrecognized ones. Previously the classifier's default for any failure reason outside its allowlist
was skip/escalate — so a provably side-effect-free failure (e.g. `cluster_fence_stale`) was skipped
every tick, indefinitely, with no per-ticket record. In the motivating 2026-08-07 incident a
fence-stale implement failure sat idle for 4.5+ hours (`"0 processed, 11 skipped"`), neither
remediated nor escalated, until a human noticed. This was the third instance of the same pattern in
24h.

This PR closes that gap in three layers:

1. **Emission site as proof of safety.** The CTL-864 fence guard bows out *before* any guarded
   side-effect, so a fresh redispatch (which bumps the cluster generation) is always safe. It now
   stamps `retry_safe: true` into the failure payload, so the classifier need not hardcode reason
   strings.
2. **Signal plumbing.** `phase-agent-emit-complete` propagates `retry_safe` from the caller payload
   onto the phase signal as `retrySafe: true`, where the recovery-pass classifier reads it off the
   signal/evidence.
3. **Generalized bounded retry.** Any retry-safe failure — the deterministic `cluster_fence_stale`
   shortcut *or* a generic unrecognized-but-safe failure — gets a bounded default retry budget
   (the shared `RECOVERY_MAX_ATTEMPTS`) with backoff via the recovery-intent ledger. Only budget
   exhaustion escalates, and the escalation names the concrete reason and carries the retry history
   so classifier coverage gaps stay visible. Unknown-and-**unsafe** failures still escalate on the
   first evaluation (existing behavior preserved).

Per ADR-023, the new retry behavior ships shadow-observable: a `recovery.would-retry` twin event and
per-ticket `recovery.skipped.<reason>` breadcrumbs make previously-invisible silent skips queryable
in Loki.

### Recovery classifier (`execution-core/recovery-reasoning.mjs`, +305/-5)

- `retrySafeBudgetDecision(ticket, { readIntentAttempts })` — the shared bounded-retry budget
  decision. Pure over an injected ledger reader (defaults to the on-disk intent ledger; injectable
  for hermetic tests); fail-open on an unreadable ledger (`attempts=0` → retry allowed).
  `attempts < MAX → { retry: true }`, `attempts ≥ MAX → { retry: false }`.
- `buildRetrySafeExhaustionExplanation(ticket, reason, attempts)` — reason-named `coverage_gap`
  escalation reusing the existing `buildEscalationPayload` tagged-union shape, so the operator inbox
  renders a real "unrecognized retry-safe failure `<reason>`" card carrying the attempts history.
- `sanitizeReasonForEvent(reason)` — slugs an arbitrary failure reason into a safe
  `[a-z0-9_-]` event-name suffix for the `recovery.skipped.<reason>` breadcrumb.
- Per-ticket `recovery.skipped.<reason>` breadcrumb for reason-bearing skips, and a
  `recovery.would-retry` twin for retry-safe redispatches (both on the deliberately-unprotected
  `recovery.*` namespace). The `cluster_fence_stale` deterministic redispatch is gated through the
  same budget so an infinitely-stale fence can't loop forever.

### Fence guard (`lib/cluster-fence-guard.sh`, +6/-1)

- The stale bow-out now emits `--payload-json '{"retry_safe":true}'` alongside the
  `cluster_fence_stale` reason.

### Signal writer (`phase-agent-emit-complete`, +11/0)

- Extracts `retry_safe` from the caller payload and persists it as `retrySafe: true` on the phase
  signal. Back-compatible: only a literal `true` propagates; a malformed/absent payload leaves the
  signal unchanged (no `retrySafe` key).

### Tests (+471)

- `recovery-reasoning.test.mjs` (+249), `fence-stale-redispatch-seam.test.mjs` (+162),
  `recovery-evidence.test.mjs` (+21), `phase-agent-emit-complete.test.sh` (+35),
  `cluster-fence-guard.test.sh` (+4). Budget/escalation decisions tested against injected ledger
  readers (hermetic).

## How to Verify It

- [x] Bun unit tests (execution-core): `bun test recovery-reasoning.test.mjs fence-stale-redispatch-seam.test.mjs recovery-evidence.test.mjs` — pass
- [x] Shell tests: `phase-agent-emit-complete.test.sh`, `cluster-fence-guard.test.sh` — pass
- [x] Back-compat: a failure with no `retry_safe` payload leaves the signal free of `retrySafe`

> ⚠️ CI `execution-core-unit-tests` currently reports one failing test —
> `CTL-1240 Phase 2 — census closures use { cache, gateway } > default stall-clear census uses { cache, gateway } via runningOpts`
> (`integration-ctl-1240.test.mjs`). This test **also fails on a clean `origin/main`** and is
> **unrelated to this change** (this PR touches neither `scheduler.mjs` nor that test). It is a
> pre-existing stale-test failure on `main` and should be fixed in its own PR.

## Related Issues/PRs

- Fixes https://linear.app/coalesce/issue/CTL-1679
- Rollout per ADR-023 (shadow → enforce).

skip ADR-023
skip CTL-1240
skip CTL-864

### DIFF
--- a/plugins/dev/scripts/__tests__/phase-agent-emit-complete.test.sh
+++ b/plugins/dev/scripts/__tests__/phase-agent-emit-complete.test.sh
@@ -780,6 +780,41 @@ else
 	assert_eq "boom" "$P_REASON" "malformed payload: base failure_reason preserved"
 fi
 
+# CTL-1679 Phase 2: --payload-json retry_safe rides into BOTH the event payload
+# AND the phase-<P>.json signal (as retrySafe). Absent --payload-json → no
+# retrySafe key (back-compat).
+echo ""
+echo "Test 49 (CTL-1679): --payload-json retry_safe → event payload + signal.retrySafe"
+fresh_env t49
+SIGNAL="${CATALYST_ORCHESTRATOR_DIR}/workers/CTL-100/phase-pr.json"
+echo '{"status":"running","ticket":"CTL-100","phase":"pr","generation":3}' >"$SIGNAL"
+"$EMIT_SCRIPT" --phase pr --ticket CTL-100 --status failed \
+	--reason "cluster_fence_stale" --payload-json '{"retry_safe":true}' >/dev/null 2>&1
+LINE=$(read_event_line)
+if [[ -z $LINE ]]; then
+	fail "Test 49: no event line emitted"
+else
+	P_RETRY=$(echo "$LINE" | jq -r '.body.payload.retry_safe')
+	P_REASON=$(echo "$LINE" | jq -r '.body.payload.failure_reason')
+	assert_eq "true" "$P_RETRY" "event body.payload.retry_safe = true"
+	assert_eq "cluster_fence_stale" "$P_REASON" "event failure_reason still carried"
+fi
+SIG_RETRY=$(jq -r '.retrySafe' "$SIGNAL")
+SIG_REASON=$(jq -r '.failureReason' "$SIGNAL")
+SIG_GEN=$(jq -r '.generation' "$SIGNAL")
+assert_eq "true" "$SIG_RETRY" "signal.retrySafe = true"
+assert_eq "cluster_fence_stale" "$SIG_REASON" "signal.failureReason preserved"
+assert_eq "3" "$SIG_GEN" "signal.generation preserved"
+
+echo ""
+echo "Test 50 (CTL-1679): absent --payload-json → no retrySafe key on signal (back-compat)"
+fresh_env t50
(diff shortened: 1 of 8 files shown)


ryanrozich:
Landed as #3835 — your commits cherry-picked onto current main with authorship preserved (2026-08-21 backlog sweep; verification + conflict-resolution rationale in the new PR's body). Review happens there.

PR #3149, opened 2026-08-08
fix(execution-core): use a valid ticket-key format in CTL-1240 stall-clear fixture

thagale:
## Changes

The Site-4 "default stall-clear census" test in `integration-ctl-1240.test.mjs` seeded its stalled signal under ticket id `"CTL-1240-STALL"`, which does not match `TICKET_KEY_RE` (`^[A-Z][A-Z0-9_]*-\d+$`) — the trailing `-STALL` suffix isn't all digits.

`defaultCollectStallClearCandidates`'s CTL-1504 debris-filter guard (`if (!isTicketKey(ticket)) continue;`) skips any worker-dir name that isn't a canonical ticket key, so the fixture was filtered out before the census ever reached the `isLinearTerminal` probe. The gateway spy was never consulted, and the test's own assertion (`expect(gateway.calls).toContain(STALL_TICKET)`) failed — a pre-existing red required check (`execution-core-unit-tests`) on `main`, blocking every PR merge in the repo.

`scheduler.mjs`'s production code is correct as-is (the gateway threading itself works). This is purely a test-fixture naming bug, most likely introduced when CTL-1504's guard was added to `defaultCollectStallClearCandidates` sometime after this test was originally written with an intentionally-descriptive-but-invalid ticket id.

Fix: renamed the fixture ticket to a canonical `TEAM-123`-shaped id (`CTL-99991240`).

## Verification
- `bun test integration-ctl-1240.test.mjs`: 3 pass / 0 fail (was 2 pass / 1 fail on `main` at `0165f78c`)
- Full `execution-core` suite (245 files, 8100 tests): failure set is identical before/after this change (minus the one fixed test) — no regressions introduced.

## Test plan
- [x] Targeted test file passes
- [x] Full execution-core suite shows no new failures vs. baseline

### DIFF
--- a/plugins/dev/scripts/execution-core/integration-ctl-1240.test.mjs
+++ b/plugins/dev/scripts/execution-core/integration-ctl-1240.test.mjs
@@ -189,7 +189,11 @@ describe("CTL-1240 Phase 2 — census closures use { cache, gateway }", () => {
   //           → spy consulted.
 
   test("default stall-clear census uses { cache, gateway } via runningOpts", () => {
-    const STALL_TICKET = "CTL-1240-STALL";
+    // CTL-1504 requires a canonical TEAM-123 key (`^[A-Z][A-Z0-9_]*-\d+$`) — a
+    // trailing letter suffix like the old "CTL-1240-STALL" is debris-filtered
+    // by isTicketKey() before defaultCollectStallClearCandidates ever reaches
+    // the isLinearTerminal probe, so the gateway spy was never consulted (CAT-62).
+    const STALL_TICKET = "CTL-99991240";
 
     writeFileSync(join(orchDir, "state.json"), JSON.stringify({ maxParallel: 1 }));
     // Seed the stalled signal — stalledReason triggers the J3 isLinearTerminal probe.


thagale:
**Superseded — CTL-1240 is already fixed on `main`.**

CAT-40's #3143 landed `STALL_TICKET = "STALL-1240"` (a valid `^[A-Z][A-Z0-9_]*-\d+$` key), which is the same fix this PR makes with `CTL-99991240`. That is why this PR now shows `CONFLICTING` — both edit the same line.

For the record, the diagnosis here was correct and mine (posted on CAT-62 earlier today) was not: `isTicketKey("CTL-1240-STALL")` is false, so CTL-1504's debris filter skips the fixture before `defaultCollectStallClearCandidates` ever reaches the `isLinearTerminal` probe. I verified this change passes 3/3 locally on a clean worktree cut from `origin/main`, and it is non-vacuous.

I also pushed an empty commit here to re-trigger the required check, because the earlier run failed on `orphan-sweep kill-gate test (CTL-1531)` — the CAT-16 load flake, unrelated to this diff — and `gh run rerun` needs admin. That retry went **green** (run 31283165811), confirming the failure was the flake and not this change.

Recommend closing as superseded by #3143. I don't have write access to close it here.

*— recovery-pass delegate, dispatched on CAT-29*

ryanrozich:
Closing as **superseded** — the fix this PR makes is already on `main`, and in the form this PR's own open review thread was asking for.

**Verified before closing:**

- This PR's entire diff is one file, `integration-ctl-1240.test.mjs`, changing `STALL_TICKET` from `"CTL-1240-STALL"` to `"CTL-99991240"` so the CTL-1504 `isTicketKey()` debris filter stops skipping the census under test. Nothing else.
- `main` already carries that same fix via **#3148**, using `PROJ-1240` / `PROJ-1241` — equally canonical, and additionally using the repository-neutral `PROJ` placeholder prefix per AGENTS.md → "Version Control". That is precisely what the unresolved P1 on this PR asks for, so the superseding version resolves it rather than leaving it open.
- #3148 also tightened the assertion beyond a key rename: it pins `calledFrom` to the exact collector frame (`defaultCollectStallClearCandidates`) instead of the module name, because `stall-janitor.mjs` contains both the J3 stall-clear census and the J4 terminal-signal GC census — so a module-level pin can be satisfied by the wrong census.
- Confirmed `integration-ctl-1240.test.mjs` passes on current `main` (3/3), and that this was the failure making `execution-core-unit-tests` red on main (tracked as CTL-1726, now Done).

Nothing from this PR is lost by closing it. The diagnosis here was correct and is what got fixed — thank you for it. Reopen if you see something in this version that #3148's does not cover.

PR #3154, opened 2026-08-08
diag: temporary CTL-1502 forward-restart CI probe (not for merge)

thagale:
Temporary diagnostic to capture the actual forward-restart hot-swap failure detail from CI. Will be closed without merging.


thagale:
Diagnostic complete — confirmed this is a test-side observer lag, not a production stop-sequence defect. Real fix follows in a separate PR.

PR #3163, opened 2026-08-08
fix(dev): CAT-23 keep dispatch ownership on live roster during outages

thagale:
8fe9a144 fix(dev): keep dispatch ownership on live roster during outages

Refs: CAT-23


ryanrozich:
Closing after verification against current main (2026-08-21 full-backlog sweep, Ryan-directed): main's CTL-1091 (resolveDispatchRoster + rate-limited outage WARN, scheduler.mjs) landed 2026-07-21, predating this PR, with the observability fix for the silent-degrade complaint. The problem this PR targets is covered — reopen with evidence if you believe the covering implementation misses your case.

PR #3167, opened 2026-08-09
fix(dev): CAT-53 detect never-adopted replica writer

thagale:
48658bfa fix(dev): detect never-adopted replica writer
6f01c12b fix(dev): read cache reconcile settings from config
8f804809 fix(dev): gate board health on ingress freshness
5c4b275d fix(dev): fail loudly on stack component startup
1a53ac4d fix(dev): reap orphaned monitor port holders

Refs: CAT-53


ryanrozich:
Closing under the cloud-native pivot (ADR-0037 merged; ADR-0039 accepted 2026-08-23 — cloud is the single writer of coordination state; runners are cloud containers; resident-daemon mechanisms are being retired, and the event log moves to the DO-outbox→R2 backbone). Ryan pre-authorized draining PRs the new direction obsoletes. Reopen if the pre-cutover daemon genuinely needs this as a stopgap. — hardens broker/cache-reconcile.mjs, the per-host cache mechanism named for retirement.

PR #3171, opened 2026-08-09
fix(dev): trust the Tailscale MagicDNS name in the monitor's reply-origin allowlist

thagale:
## Summary
- Every reply POST in the orch-monitor UI 403s with "cross-origin reply rejected" on every host in a Tailscale-accessed fleet, identically.
- Root cause: `buildTrustedOrigins()` derives "own names" from `os.hostname()` and the macOS Bonjour name, but never the Tailscale MagicDNS name (`<host>.<tailnet>.ts.net`) — which is how operators actually browse a Tailscale-fronted fleet.
- Fix mirrors the existing `bonjourName()` pattern: best-effort, memoized shell-out to read `Self.DNSName` from `tailscale status --json`, trusted on the bound port, gated to a wildcard bind like every other own-name source.

## Test plan
- [x] New unit tests for `tailscaleMagicDnsName()` + `buildTrustedOrigins()` injection seam (5 new tests, 74/74 pass)
- [x] Full `orch-monitor` suite: 6752 pass, 9 pre-existing/unrelated failures confirmed present on unmodified `main` too (env-dependent `catalyst-monitor.sh` + `board-data-repo-for` tests)
- [x] `tsc --noEmit` clean
- [x] Live integration test (`inbox-conversation-endpoints.test.ts`) confirms the real running server's trusted-origin set now includes the MagicDNS origin, while still correctly rejecting `evil.example` and `null`

Already merged to thagale/catalyst#61.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016QREzicxX7ZwQup3NAfs7A


ryanrozich:
Landed as #3830 — your commits cherry-picked onto current main with authorship preserved (2026-08-21 backlog sweep; verification + conflict-resolution rationale in the new PR's body). Review happens there.

PR #3147, opened 2026-08-08
docs(dev): document CAT-60 publish preflight

thagale:
45f894c6 docs(dev): document CAT-60 publish preflight
3de981c7 feat(dev): add CAT-60 publish capability preflight
0bd99f03 fix(dev): route CAT-60 pushes through configured remote

Refs: CAT-60

PR #3177, opened 2026-08-09
fix(dev): CAT-128 stub the live skills-dir probe in doctor install tests

thagale:
## Why

`execution-core-unit-tests` is **red on `main`** (run [31294460725](https://github.com/coalesce-labs/catalyst/actions/runs/31294460725), sha `12c454ff`), so every open PR inherits a failing required check and cannot merge. Six are blocked on it today: #3140, #3141, #3147, #3156, #3166, #3170.

## Root cause

The `skills-dir-plugins` check added to the install rubric in #2664 probes a real `~/.claude` tree. Four `installChecksForClass` / `runDoctor({ profile: "install" })` tests never stubbed it, so their verdict depends on whether the host happens to have the catalyst plugin symlinks present. On a CI runner (and any clean checkout) the probe FAILs and takes the suite down with it:

```
fail  skills-dir-plugins :: catalyst plugins do not load cleanly in-place via
      user-scope skills-dir (~/.claude/skills/catalyst-pm is missing; ...)
```

This is a **test-hermeticity defect, not a product defect** — `checkSkillsDirPlugins` is behaving correctly.

## Fix

`installChecksForClass` already exposes a `skillsDirCheck` seam for exactly this reason, and `runDoctor` already forwards it. A sibling test at doctor.test.mjs:4853 uses it, with a comment explaining why. These four tests simply omitted it — so this stubs them the same way. No product code changes.

Three of the four were failing. The fourth (developer post-install) passes today only by environmental luck and carried the identical latent defect, so it is stubbed too rather than left as a future flake.

## Verification

```
bun test plugins/dev/scripts/execution-core/doctor.test.mjs
  412 pass / 0 fail    (before: 409 pass / 3 fail)
```

Refs CAT-62, CAT-131, CAT-135.

### DIFF
--- a/plugins/dev/scripts/execution-core/doctor.test.mjs
+++ b/plugins/dev/scripts/execution-core/doctor.test.mjs
@@ -4934,6 +4934,9 @@ describe("installChecksForClass — the focused post-install verification (CTL-1
         hasStackAgent: true,
         hasUpdaterAgent: true, // the two-puller hazard
         pluginPullOwner: "broker",
+        // Hermeticity: the real skills-dir probe reads a live ~/.claude tree, which a
+        // unit test must not depend on (its own coverage lives with checkSkillsDirPlugins).
+        skillsDirCheck: () => [{ name: "skills-dir-plugins", status: "pass", detail: "stubbed" }],
       }),
       log: () => {},
     });
@@ -4947,6 +4950,8 @@ describe("installChecksForClass — the focused post-install verification (CTL-1
       hasStackAgent: true,
       hasUpdaterAgent: false,
       pluginPullOwner: "broker",
+      // Hermeticity: stub the live ~/.claude skills-dir probe (see checkSkillsDirPlugins).
+      skillsDirCheck: () => [{ name: "skills-dir-plugins", status: "pass", detail: "stubbed" }],
       log: () => {},
     });
     expect(code).toBe(0);
@@ -4959,6 +4964,8 @@ describe("installChecksForClass — the focused post-install verification (CTL-1
       hasStackAgent: false,
       hasUpdaterAgent: true,
       pluginPullOwner: "updater",
+      // Hermeticity: stub the live ~/.claude skills-dir probe (see checkSkillsDirPlugins).
+      skillsDirCheck: () => [{ name: "skills-dir-plugins", status: "pass", detail: "stubbed" }],
       log: () => {},
     });
     expect(code).toBe(0);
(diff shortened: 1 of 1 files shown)


thagale:
Flagging this as the **keystone PR** for the current board, from a whole-board recovery sweep.

`main` has been red on `execution-core-unit-tests` since `12c454ff` (#2664). Because `pull_request` CI runs against `refs/pull/<n>/merge`, that red is inherited by every open PR — **12 open PRs are currently failing that check**, and for most of them it is the *only* thing failing:

```
#3140 #3141 #3146 #3147 #3156 #3160 #3163 #3166 #3167 #3170 #3175 #3176
```

None of those authors can fix it on their own branch. This PR is the fix, and it is `MERGEABLE`/`CLEAN` — merging it should let all 12 go green on their next run with no rebase needed, since they pick the fix up through the merge ref automatically.

Independent verification on a clean checkout of main (`7988f2fe`), reached separately before finding this PR:

| | Result |
| --- | --- |
| before — `bun test doctor.test.mjs` | 409 pass / **3 fail** |
| after — `bun test doctor.test.mjs` | **412 pass / 0 fail** |
| after — full CI "Run stable unit tests" list (172 files) | **6020 pass / 0 fail** |

I had opened #3178 with the same fix before seeing this one; it is closed in favour of this PR, which is also strictly broader — it stubs the seam at 4 sites rather than 3. The 4th, `PASSes a correctly-provisioned developer post-install`, passes today only because the skills-dir check WARNs rather than FAILs on a developer class, so it is a latent failure this PR closes too.

One caveat worth knowing when reading this PR's checks: `forward-restart.test.sh` (CTL-1502) is an early, fail-fast step in the same job and flakes intermittently ("old forwarder pid N still alive after restart"). When it loses, it skips the following **17** steps including `Run stable unit tests` — so a red `execution-core-unit-tests` here may be that race rather than a real regression. Tracked as CAT-70; it red-lit #3178 for exactly that reason.

ryanrozich:
Closing as **superseded by #3180**, which fixed the same breakage and is now merged.

Both PRs correctly diagnosed it, and the diagnosis was right: the `skills-dir-plugins` check I added to the install rubric in #2664 probes a live `~/.claude` tree, so the install tests graded the host they ran on. It passed on my machine (which has the symlinks) and went red in CI. **That was my break — thank you for catching and fixing it.**

I merged #3180 rather than this one purely on factoring: it extracts a single `passingSkillsDirCheck` helper instead of inlining the stub at each site, and its comment explains *why* the seam matters (for `class=worker` a missing symlink is a hard FAIL, while developer/monitor only WARN — so any test asserting on the aggregate FAIL count must inject it). Same fix, less duplication.

Verified before merging, rather than trusting the green check: under a hermetic `HOME`, `main` fails 3 tests and #3180's branch passes 412/412.

Nothing in this PR is lost by closing it.

PR #3187, opened 2026-08-10
docs(cluster-onboarding): correct stale roster mechanism, add real gotchas

thagale:
## Summary
- The Quick Start's `--seed`/`--token` flag syntax was stale — the script takes `CATALYST_SEED`/`CATALYST_JOIN_TOKEN` env vars, not CLI flags — and cloning just `catalyst-join.sh` is insufficient (it needs its sibling scripts: `provision-thoughts.sh`, `install-cli.sh`, `setup-plugin-source.sh`).
- The Activation section documented only one of the two roster mechanisms the doc's own precedence list already promises (`resolveClusterHosts()`: cluster-repo → static-roster → single-host). Restructured into **Option A** (static-roster escape hatch) and **Option B** (committed `catalyst-cluster` roster), each with its own real steps, converging on a shared "after either option" tail.
- Added the webhook-wiring re-run step for Option B — Stage-0 join deliberately skips webhook wiring (`stage0-roster-guard`), and re-running the join with fresh credentials after roster activation is a real, easy-to-miss step (`catalyst-doctor` FAILs `webhook-ingestion` until it's done).
- Added real onboarding gotchas hit during actual multi-node work: bash 3.2 on a fresh macOS node breaking newer-bash-only syntax, a node's runtime identity (baked into the launchd plist) not self-correcting after a `catalyst cluster rename`, plist env changes needing a full stop rather than `launchctl kickstart`, Codex needing its own separate per-node auth, webhook ingestion not propagating via the join bundle, thoughts org/convention needing to be explicit for a forked install, and Layer-2 config file permissions drifting back to 644 after certain rewrites.

Resolved several conflicts against current `main` (the doc has moved independently since this was authored) — kept both sides' real content where they covered different ground, rather than dropping either.

## Test plan
- [ ] Read-through for accuracy against current `main`'s actual mechanism
- [ ] Maintainer review

https://claude.ai/code/session_01MzRkTyQ7uZSLCwzkWPeqbm


ryanrozich:
Landed as #3832 — your commits cherry-picked onto current main with authorship preserved (2026-08-21 backlog sweep; verification + conflict-resolution rationale in the new PR's body). Review happens there.

PR #3188, opened 2026-08-10
feat(dev): give linearis its own dedicated app-actor identity

thagale:
## Summary

`linearis` (the CLI behind nearly all fleet Linear reads/writes) authenticates as the operator's personal Linear account for two specific automated call sites, so their traffic shares the operator's personal rate-limit bucket. This gives it a dedicated, non-personal identity instead, following the exact same pattern already established for the orchestrator/worker app-actors.

Verified directly (reading the actual code, not assuming) that execution-core and broker were *already* isolated from the operator's personal quota — they inherit the orchestrator's app-actor token at daemon startup. The two real gaps were a LaunchAgent sweep with no app-actor auth at all, and a dispatcher whose spawned workers don't inherit a plain env export.

## Changes

1. **New secret-contract row** — `linear-linearis-actor`, same shape as the existing `linear-orchestrator-actor`/`linear-worker-actor` rows (JS registry + bash mirror + parity tests).
2. **Parameterized the shared mint function** — `linear_app_actor_auth` used to be hardcoded to always mint the orchestrator's credentials. It now takes optional `[secret-id] [display-name]` params, defaulting to the exact old values, so the three existing 2-arg callers are byte-for-byte unaffected (explicit regression test included).
3. **Wired the two real gaps**:
   - `dependabot-escalate.sh` — had zero app-actor auth; now mints at script start.
   - `phase-agent-dispatch` — threads the token through the same `--settings` JSON channel already used for OTEL/context vars (a plain env export gets silently dropped across the `claude --bg` RPC boundary).
   - Both fail-open: a failed mint just leaves things exactly as they behave today, nothing blocks.
4. **Regression tests, no production change** — locking in that this new identity stays invisible to the two places that currently decide "is this write bot-authored," so writes through it keep being treated as human-authored, matching today's actual behavior.

## Testing

Full TDD throughout, verified against both `origin/main` and `thagale/catalyst` (already merged there as #69, all CI green). All green except one transient CI flake in an unrelated file (`orphan-sweep.sh`'s kill-gate suite — process-signal-timing tests, confirmed reproducible-clean on retry, this diff never touches that file).

```
secret-contract.test.mjs + linear-comment.test.mjs:  141/141
linear-app-actor.test.sh:                             39/39
dependabot-escalate.test.sh:                           4/4
secret-contract-parity.test.sh:                      181/181
catalyst-secret-contract.test.sh:                    115/115
phase-agent-dispatch.test.sh:                        321/322 (1 pre-existing failure, unrelated file/logic)
daemon.test.mjs:                                     180/181 (1 pre-existing failure, unrelated logic)
```


ryanrozich:

PR #3204, opened 2026-08-10
feat(dev): CAT-168 persist the relations cache to disk so it survives a daemon restart

thagale:
Mirror of thagale/catalyst#76 (merged). Opened per this fork's dual-PR policy — not intended to be merged upstream by us.

## Summary
- `linear-cache.mjs`'s relations store was in-memory only and wiped on every daemon restart, with no durable tier behind it (Linear's webhook subscriptions never carry `IssueRelation` events). Every restart forced a full live re-fetch of every ticket's relations.
- Adds `linear-relations-store.mjs`: a `bun:sqlite`-backed L2, deliberately a separate db file from the webhook-fed replica.
- The existing in-memory Map stays the L1 (hot per-tick reads, unchanged latency); the store is consulted only on a cold Map miss.
- The ticket-STATE store is out of scope — it already sits in front of a durable replica tier.


ryanrozich:

PR #3205, opened 2026-08-10
fix(dev): CAT-47 commit the audit-comment dedup latch only on confirmed delivery

thagale:
## Summary

PR #3169 (CAT-47) merged without the phase-review remediation commit that fixed its own
HIGH-severity finding. This lands that commit against current `main`.

The gap surfaced during CAT-47's teardown open-PR/Done judgment: `main` today has 8 tests in
`recovery-orphan-stale-e2e.test.mjs` and no delivery gate, while the reviewed remediation
(`68f65b6e`, authored 2026-08-08) has 10 and the gate. It was never pushed to the branch that
became #3169.

## The bug

`reasoningRecoveryPass` committed the audit-comment dedup latch (`commitFixCommentHashFn`) on **any
non-throwing** `postComment` return. But `defaultPostComment` fails **closed without throwing** — a
non-zero `linear-comment-post.sh` exit returns `{ok: false}`, and the scheduler does not override
it. A Linear outage or auth failure therefore committed the hash and suppressed that audit comment
**permanently**.

This directly contradicts the contract `docs/architecture.md` already states on `main` (line 433):

> Audit-comment hashes are committed only after successful delivery, so an outage leaves the comment
> eligible for retry while delivered duplicate content is suppressed.

The branch's one delivery-failure test used a **throwing** stub, so the production failure shape was
uncovered.

## Changes

- `recovery-reasoning.mjs` — only an explicitly non-failing result counts as delivery. A seam that
  returns nothing is still treated as delivered (for those, a throw is the failure signal), so
  injected test seams keep working.
- Isolated the latch write from the post: a failure *there* means the comment **was** delivered but
  the latch did not land, and is now logged as a latch-write failure instead of being misattributed
  to the post.
- Two pinning tests: a non-throwing `{ok:false}` post re-posts next tick; an `{ok:true}` post
  commits the hash and suppresses the duplicate.
- `docs/architecture.md` — corrected the CAT-47 section, which claimed Pass 0u builds its registry
  from the shared `unstuckSeamDeps` bundle. It does not: Pass 0u still constructs its own equivalent
  closures inline; the bundle serves Pass 0r's capability-checked fallback only.

## Verification

```
$ bun test recovery-orphan-stale-e2e.test.mjs recovery-fix-backoff.test.mjs \
           unstuck-act-seams.test.mjs unstuck-orphan-merge.test.mjs
 69 pass, 0 fail — 137 expect() calls, 4 files
```

Non-vacuity checked: reverting only `recovery-reasoning.mjs` to `origin/main` makes
`a non-throwing {ok:false} post does not commit the dedup hash` fail (`Expected: 2, Received: 1`);
restoring it turns it green.

Refs: CAT-47

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -443,11 +443,14 @@ The behavior is gated by `CATALYST_RECOVERY_PASS` (off by default); shadow mode 
 
 ### Orphan-stale merged-PR reconciliation (CAT-47)
 
-Pass 0r and Pass 0u share the same production act-seam dependencies. `runTick` constructs the
-PR-state resolver, background-job liveness probe, stall clearer, and status writer once; Pass 0u
-uses them to build its act registry, while Pass 0r receives both that registry and the raw bundle
-for capability-checked fallback construction. An injected partial registry therefore falls back to
-real dependencies instead of either using inert defaults or failing as unavailable.
+Pass 0r and Pass 0u act on the same production act-seam dependencies. `runTick` builds a
+`unstuckSeamDeps` bundle — PR-state resolver, background-job liveness probe, stall clearer, status
+writer — and threads it to Pass 0r, which receives both that bundle **and** Pass 0u's prebuilt act
+registry. `defaultInvokeSeam` then selects by capability: an absent or partial registry falls back
+to a registry rebuilt from the bundle's real dependencies instead of either using inert defaults or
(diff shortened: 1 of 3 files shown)


ryanrozich:
Landed as #3831 — your commits cherry-picked onto current main with authorship preserved (2026-08-21 backlog sweep; verification + conflict-resolution rationale in the new PR's body). Review happens there.

PR #3240, opened 2026-08-11
CAT-154: Developers should see deterministic doctor.test.mjs results

thagale:
## Summary

Makes the execution-core doctor tests deterministic across hosts and full-suite ordering by isolating the skills-directory check from live `~/.claude` state. The change is test-only: it injects the existing `skillsDirCheck` seam at seam-configuring call sites, pins `CLAUDE_CONFIG_DIR` beneath the hermetic test root, and adds guards that make future seam omissions fail locally.

## Changes Made

- Injected `passingSkillsDirCheck` wherever `installChecksForClass` or install-profile `runDoctor` tests configure seams.
- Added a source guard that scans call sites both before and after its own delimited body, validates its region markers, identifies top-level option arguments, and trips on uncovered seam-configuring calls.
- Pinned and recorded the test preload's Claude configuration directory under `CATALYST_HERMETIC_DIR`.
- Added preload tests covering path isolation, guaranteed absence, live `claudeConfigDir()` resolution, and agreement between the live pin and its stable record.
- Documented the pin's precise scope: other doctor readers that use the real home directory remain outside this test-only change.

## How to Verify It

- [x] Targeted execution-core tests pass: `cd plugins/dev/scripts/execution-core && bun test doctor.test.mjs test-setup.test.mjs` — 424 passed, 0 failed.
- [x] Phase verification found the branch and `origin/main` had the same pre-existing full-suite failure set.
- [x] Determinism checks passed across isolated runs and alternate `HOME` / `CLAUDE_CONFIG_DIR` inputs.
- [x] Mutation checks proved the seam guard fails when a seam or region marker is removed and when an uncovered call is introduced.
- [ ] CI checks complete on the pull request.

## Verification Notes

The pre-remediation verification recorded 56 pre-existing execution-core full-suite failures on both the branch and the `origin/main` control, with no branch-only failures. The final remediation commit strengthens the source guard and its tests without changing application code.

## Reviewer Notes

This PR intentionally changes only test infrastructure. Follow-up production or broader-suite work is tracked separately in CAT-247, CAT-248, CAT-249, CAT-250, and CAT-251.

## Related Issues/PRs

- Fixes https://linear.app/coalescelabs/issue/CAT-154

## Changelog Entry

No user-facing changelog entry; test determinism and test-harness isolation only.

skip CAT-247
skip CAT-248
skip CAT-249
skip CAT-250
skip CAT-251


ryanrozich:
Closing after verification against current main (2026-08-21 full-backlog sweep, Ryan-directed): main's doctor.test.mjs already wires passingSkillsDirCheck at every call site + hermetic dir pinning in test-setup.mjs. The problem this PR targets is covered — reopen with evidence if you believe the covering implementation misses your case.

PR #3242, opened 2026-08-11
fix(dev): reap an orphan squatting the monitor's service port

thagale:
Upstream contribution (courtesy PR — not intended to be merged by this fork's operator). See thagale/catalyst#97 for the corresponding merged fix.

CAT-53: an orphaned orch-monitor from an earlier stack generation (PPID 1, no tracked pidfile) could keep squatting the service port indefinitely while catalyst-stack reported "stack up" underneath it. Adds orphan detection/reap before start, and gates the "stack up" banner on a real liveness check instead of printing unconditionally.


ryanrozich:

PR #3269, opened 2026-08-12
fix(dev): CTL-1791 — delete claim.mjs dispatch-claim, the high-water+1 generation the CTL-736 fence forbids

ryanrozich:
Part of CTL-1791.

## Why

`claim.mjs`'s `dispatch-claim` subcommand has **no production caller** — but it is **armed, not inert**. It implements the high-water+1 generation algorithm that `phase-agent-dispatch:678-704` names in its own comments as the live CTL-736 double-spawn bug, **and the test asserted that behaviour as correct**. Deleting the subcommand without the test would leave a test certifying the bug; deleting neither leaves a loaded gun with a friendly usage string.

## What

- Removes the `dispatch-claim` case arm and its usage token.
- Replaces the test that codified the forbidden behaviour with a guard test asserting the subcommand is **rejected** (exit 1), creates no claim file, and is absent from usage.
- Corrects two **factually false** doc comments found in the process: `claim.mjs`'s header claimed it is imported by `recovery.mjs` and shelled into by two bash scripts (none of which is true — `recovery.mjs` imports `cluster-claim-sync.mjs`/`fence-event.mjs`, and both bash scripts reimplement the semantics), and `currentGeneration`'s doc actively recommended *"A spawn-path dispatcher claims this + 1"* — the exact derivation this deletion exists to forbid. Deleting the subcommand while leaving comments recommending it is how it gets reintroduced.

## Scope correction — the rest of CTL-1791 is NOT in this PR

The ticket also proposed deleting the ADR-011 SQLite archive index. **That premise was proven false during implementation and the item is blocked**, not skipped:

- `~/catalyst/archives/<TICKET>/` (written by `phase-teardown`) and `~/catalyst/archives/<orchId>/` (written by `catalyst-archive.ts sweep`) are **different namespaces sharing a parent directory**. mini's 91 populated dirs are ticket-keyed. Deleting `catalyst-archive.ts` would remove the entire orchId-keyed artifact-persistence system for the `catalyst-legacy` path, which is installed and documented as the runtime fallback — not an unused index.
- The proposed replacement teardown gate (`.teardown-complete`) is present on only **2 of 91** dirs, and those dirs are ticket-keyed while `/catalyst-dev:teardown` takes an orchId. It would swap an always-failing gate for a differently-always-failing one that *looks* correct, on a skill guarding destructive worktree removal.

The ticket has been updated with this correction.

## Verification

`claim.test.mjs` 20→21 pass. `grep -rn dispatch-claim` over the tree returns only the two edited files. Adversarially reviewed against the diff: **VERDICT: SAFE-TO-MERGE**.

https://claude.ai/code/session_0144RKpHEbUrj5UHTF5JeAjk

### DIFF
--- a/plugins/dev/scripts/execution-core/claim.mjs
+++ b/plugins/dev/scripts/execution-core/claim.mjs
@@ -10,18 +10,24 @@
 //   • claimPhase(orchDir, ticket, phase, generation) — open(O_CREAT|O_EXCL) of
 //     ${orchDir}/workers/<ticket>/<phase>.claim.<generation>. Exactly one
 //     caller wins per generation; concurrent same-generation callers collide.
-//   • The generation is a monotonic FENCING TOKEN. A fresh dispatch claims
-//     gen=currentGeneration+1 (1 when nothing is held); a revive (sequential,
-//     post-death) claims the next generation — a NEW filename, so O_EXCL
-//     succeeds for it regardless of whether the dead generation's claim file is
-//     still on disk.
+//   • The generation is a monotonic FENCING TOKEN, derived from the SIGNAL, not
+//     from the live claim set: a fresh dispatch (no signal) claims 1; a revive
+//     claims signal.generation + 1 — a NEW filename, so O_EXCL succeeds for it
+//     regardless of whether the dead generation's claim file is still on disk.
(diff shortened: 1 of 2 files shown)


ryanrozich:
@codex review

PR #3216, opened 2026-08-11
test(dev): CAT-135 add injectable doctor host-state seams

thagale:
## Summary

Adds injectable seams for execution-core doctor checks that inspect host state, while preserving the existing live probes as production defaults. A frozen registry now documents which seams belong to the activation and install rubrics and supports contract tests across worker, developer, and monitor node classes.

## Changes Made

- Added `HOST_STATE_SEAMS` as the frozen registry for `skillsDirCheck`, `pluginSourceFreshnessCheck`, and `staleLockCheck`.
- Wired the activation rubric to honor all three injected checks by function identity.
- Kept the install rubric's skills-directory check injectable and verified its failure propagation.
- Added registry, identity, live-default, host-environment independence, and load-bearing mutation coverage.
- Updated execution-core test setup commentary for the host-state isolation boundary.

## How to Verify It

- [x] Doctor tests pass: 425/425.
- [x] All six suites importing `doctor.mjs` introduce no new failures: 557/560, with three pre-existing failures also present at the merge base.
- [x] The explicit CI suite introduces no new failures: 6,193 passing / 30 failing on the branch versus 34 failing at the merge base.
- [x] Security scan found no added secrets or dependency changes.
- [x] Changed executable lines are covered, and adversarial mutations make the new guards fail.
- [ ] Type checking is not applicable to the changed `.mjs` files.
- [ ] No configured lint gate applies to these `.mjs` files.

## Reviewer Notes

The verification phase assessed regression risk as 1/10. Production behavior is unchanged unless a caller supplies one of the new whole-check seams.

## Related Issues/PRs

- Refs: CAT-135

## Changelog Entry

Test execution-core doctor host-state checks through explicit injectable rubric seams.

PR #3273, opened 2026-08-12
test(dev): CAT-179 — host-state seam guardrails (CAT-135 deferred findings)

thagale:
Addresses the three review findings CAT-135 deferred onto CAT-179.

- **Finding 3** — the `HOST_STATE_SEAMS` live-check assertion joined the whole suite before matching, so any unrelated thunk mentioning the identifier satisfied it. Now call-shaped (`checkStaleLock(`, CTL-1415's proven form) and counted per-thunk via `.filter().toHaveLength(1)`.
- **Finding 2** — the registry is opt-in and `runDoctor` spreads `opts` opaquely, so an unregistered seam is a silent no-op. Adds `doctor-host-state-registry.test.mjs`: a source scan (`readFileSync`, not `toString()` — bun transpiles) that fails when a `check*` whose own body calls `homedir()` is neither registered nor in the two-directional `KNOWN_UNREGISTERED` allowlist. Wired into the CI file list. Registry header comment softened to state its actual scope.
- **Finding 1** — the `test-setup.mjs` deferral comment implied the deferred HOME pin *caused* 26 failures. Reworded to the measured facts (30 under real HOME, 26 under empty HOME, the 26 a strict subset, none in `doctor.test.mjs`; host-specific, not re-measured) and corrected the remedy: under bun an in-process `process.env.HOME` change moves neither `os.homedir()` nor child env, so a pin here could not reach `doctor.mjs`'s 20 `homedir()` sites. `CLAUDE_CONFIG_DIR` *would* take effect and is labelled untaken scope.

## Verification

`bun test doctor.test.mjs doctor-host-state-registry.test.mjs` (phase env stripped per CAT-170) → **429 pass, 0 fail**.

Each guard is proven by mutation, then reverted:

| Mutation | Expected failure | Result |
|---|---|---|
| Unwire `staleLockCheck`, add decoy thunk mentioning the identifier | tightened assertion + CTL-1415 | both fail ✓ |
| Add a bogus `homedir()`-reading `check*` | *accounts for every homedir()-reading check* | fails ✓ |
| Drop `checkLogShipper` from the allowlist | same + *pins the currently unregistered set* | fails ✓ |
| Add `checkNotAThing` to the allowlist | *has no stale allowlist entries* | fails ✓ |

## Base

The plan proposed stacking on `fork/CAT-135`. This branch instead carries CAT-135's diff as commit `311c2122`, verified byte-identical to `fork/CAT-135@8a738344` across all three files — so this is a **standalone PR against `main` that contains #3216's changes**, not a stacked PR. #3216 should be closed as superseded, or this PR rebased to drop `311c2122` once #3216 merges.

No production behavior changes: two comment blocks, one CI file-list line, and test code.


ryanrozich:
Closing after verification against current main (2026-08-21 full-backlog sweep, Ryan-directed): built on #3216's superseded baseline; the PR's own body cites CTL-1415 as the proven form, which is on main. The problem this PR targets is covered — reopen with evidence if you believe the covering implementation misses your case.

PR #3281, opened 2026-08-12
fix(dev): CAT-73 — resolve replica token presence from the writer's environment

thagale:
## Summary

Closes the four accuracy gaps deferred from CAT-35 / PR #3144 round-2 review. One is behavioral and
load-bearing; three are single-location documentation corrections whose contradicting authority was
already in-repo.

**The core defect.** `checkCloudSync` (`execution-core/doctor.mjs`) read the cloud-token value from
`process.env`. The `catalyst doctor` invocation chain sources nothing, while `cloud-sync/launch.sh`
— the launcher that actually starts the replica writer — sources `~/.config/catalyst/cluster.env`
and `~/.config/catalyst/cloud-sync.env`. A host that followed the canonical runbook therefore had a
healthy, authenticated writer reported as `replica-token WARN` / `replica-tier … partially
configured`.

### Approach — probe, don't source

Rather than adding a sixth hand-rolled copy of the source-both-files chain (or injecting the secret
into the doctor process), this reuses the existing launchd-faithful presence probe
`lib/cloud-sync-token-probe.sh` and injects only its answer.

|  | source in `catalyst-doctor` | inject `env` in JS | **probe (chosen)** |
| --- | --- | --- | --- |
| Secret enters the doctor process | yes | yes | **no — only a boolean crosses** |
| Models the writer's real (launchd) view | no | no | **yes** |
| Reuses a tested primitive | no | no | **yes** |

`catalyst-doctor` (the bash entrypoint) is unmodified.

### Deliberate behavior change

A third verdict state falls out: token present in the **invoking shell but not in the writer's
environment**. That previously reported PASS — a false green, since the writer genuinely cannot
authenticate. It is now a WARN naming the precise remedy. Probe *unavailability* is likewise an
explicit, message-visible degraded state rather than a silent fall-through to `process.env`
(honouring the CAT-154 learning that an injectable seam's omission should fail loudly).

## Documentation corrections

- `website/.../reference/configuration.md` — the canonical runbook now names the resolved
  cloud-token variable and cross-links the token-name escape hatch.
- `docs/linear-replica.md` — the seed condition requires the `sync_meta` cursor and routes the seed
  check through `verify-cloud-sync`; `service_namespace` is correctly named a stream label (it
  previously claimed `service_name` was the only one).

## Verification

- `execution-core/__tests__/doctor-replica.test.mjs` — **27 pass / 0 fail** (454 assertions)
- `__tests__/cloud-sync-token-probe.test.sh` — **18 pass**
- `__tests__/docs-drift-references.test.sh` — all 8 new CAT-73 assertions pass

> Two assertions in `docs-drift-references.test.sh` fail on this branch, both **pre-existing on
> `main`** and untouched here: they reference
> `website/src/content/docs/reference/setup-health-check.md`, which does not exist on `main` either.
> Filed separately rather than fixed in-scope.

Plan: `thoughts/shared/plans/2026-08-11-cat-73.md`

### DIFF
--- a/docs/linear-replica.md
+++ b/docs/linear-replica.md
@@ -17,8 +17,15 @@ the token FIRST, activate the writer, wait for a verified seed, and only then fl
    `~/.config/catalyst/cloud-sync.env` (`chmod 600`).
 2. **Activate the writer**: `catalyst-stack adopt-cloud-sync`. Provisioning the token does not
    itself install or start the supervised writer.
-3. **Wait for a verified seed** — `catalyst doctor`'s `replica-fresh` PASS, or
-   `sqlite3 ~/catalyst/catalyst-replica.db 'SELECT COUNT(*) FROM issues'` > 0.
+3. **Wait for a verified seed** — `catalyst-stack verify-cloud-sync` (this is the authority: it
+   checks the schema, issue rows, a fresh writer lock, **and** a non-empty seed cursor). Equivalent
+   manual check — the row count alone is **not** sufficient, because a database with `issues` rows
+   but no `sync_meta` cursor is mid-reseed and every read still falls back to `linearis`:
+
+   ```bash
+   sqlite3 ~/catalyst/catalyst-replica.db \
+     "SELECT 1 FROM sync_meta WHERE key='cursor' AND value<>'' LIMIT 1;"
+   ```
 4. **Then set `CATALYST_LINEAR_REPLICA=on`** and restart execution-core on a worker — an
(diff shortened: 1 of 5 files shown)


ryanrozich:
Closing after verification against current main (2026-08-21 full-backlog sweep, Ryan-directed): superseded by CTL-1616's secret-contract resolution (resolveNodeCloudTokenEnv + shadow-diff) which replaces the raw process.env read this fixes. The problem this PR targets is covered — reopen with evidence if you believe the covering implementation misses your case.

PR #3287, opened 2026-08-12
feat(dev): CAT-3 expose fence-suppressed escalations

thagale:
Courtesy PR — mirrors a fix already merged on our fork (thagale/catalyst#83). Not intended to be merged here; opened for upstream visibility per our dual-PR convention.

Fixes `defaultEscalate()` in `stale-pr-rescue-timer.mjs` (and the equivalent scheduler.mjs terminal-sweep/dependency-cycle sites via the already-open CAT-173 courtesy PR #3241) so a zombie-guard-suppressed escalation write emits an observable `escalation.fence-suppressed` event instead of silently vanishing.

Built on top of CAT-173 (#3241) since this fix depends on its fence-standoff retry contract, which hasn't landed on `main` here yet.

PR #3288, opened 2026-08-12
fix(dev): CAT-219 window fence suppression events

thagale:
Courtesy PR — mirrors a fix already merged on our fork (thagale/catalyst#101). Not intended to be merged here; opened for upstream visibility per our dual-PR convention.

Adds a delegate-queue ceiling fallback to `defaultEscalate()`: when the delegate queue is full, falls back to a plain needs-human label instead of dropping the escalation.

Built on top of CAT-3 (#3287) and CAT-173 (#3241), which this depends on.


ryanrozich:

PR #3297, opened 2026-08-12
fix(dev): CAT-223 terminally-stalled worker directories should not keep tickets stuck in-flight

thagale:
Courtesy PR — mirrors a fix already merged on our fork (thagale/catalyst#89). Not intended to be merged here; opened for upstream visibility per our dual-PR convention.

A terminally-stalled worker directory no longer counts as in-flight, so the scheduler can re-dispatch or explicitly skip it instead of the board silently freezing with free slots and zero dispatches. Adds `dispatch-skip.mjs` (named skip reasons) and `stalled-repull.mjs` (re-pull budget).

Built on top of CAT-3 (#3287) and CAT-173 (#3241), which this depends on.


ryanrozich:

PR #3300, opened 2026-08-12
feat(dev): CTL-1790 PR-2 — three-valued outcome verdict (foundation, NOT wired)

ryanrozich:
> **DRAFT — do not merge yet.** This module has **no production caller**. Merging it as-is would ship a feature that has never produced an output, which is precisely what this project's acceptance test forbids. It merges once PR-3 wires it in shadow mode.

The structural decision behind the CTL-1790 inversion, extracted so it can be reviewed on its own after three blocked rounds.

## Why three values

Rounds 1–3 all died on the same thing: **a boolean probe conflates "I checked, the work is absent" with "I could not check."** Every probe in `work-done-probes.mjs` returns `false` for both, so inverting on `false` manufactures terminal failures out of GitHub outages, unresolvable worktrees, absent probes, and circular probes.

```
VERDICT = { DONE, NOT_DONE, UNKNOWN }
```

and the entire safety property is one rule:

> **UNKNOWN never writes `failed`.**

On UNKNOWN the caller reproduces today's behavior byte-for-byte (the flip still writes `done`) and emits telemetry. This mirrors the house rule already stated verbatim at `recovery.mjs:3027-3028` for the freshness gate.

`shouldWriteFailed` tests **positively** for `NOT_DONE` rather than `!== DONE`, so a verdict value added later cannot silently become a reason to fail work.

## The allowlist is a safety property, not an oversight

`INVERTIBLE_PHASES` admits eight phases. The two post-PR phases are exempt, for two *different* reasons:

- **`teardown`** has **no probe at all** (verified by executing `hasProbe` over all ten canonical phases, with the negative control `hasProbe("not-a-phase") === false`) **and** is the sole gate for all three Linear-Done writers. A fail-closed inversion there writes `failed`, `signals["teardown"]` never reads `done`, and **all three go dark at once** — PR merged, ticket never Done, worktree never removed. `isTicketTerminalOrMerged` would then likely *suppress* the alarm, since it returns `{terminal:true, reason:"pr-merged"}`. A silent permanent stall behind green health signals — **exactly the round-2 blocker**.
- **`monitor-deploy`** has a probe, but it reads the **same inode its writer writes** (the CTL-701 dual-use signal file). Circular by construction — no replay can exist in which probe-input and write-target are independent.

An **allowlist** (not a denylist) makes both structurally unreachable, and a phase added to the pipeline later is exempt until somebody deliberately admits it.

## Blast radius: 1 of 31, not 15 of 31

The live `16 declared / 15 fabricated` was never decomposed before. It is:

| ticket | advances | declared | fabricated |
|---|---|---|---|
| CTL-1774 | 9 | **9** (full pipeline) | 0 |
| CTL-56 | 22 | 7 | **15** |

All 15 sit on **two edges of one ticket** — 14× `monitor-merge → monitor-deploy` (13 of those a replay burst inside 55 s, re-reading one stale file) and 1× `monitor-deploy → teardown`.

Under this design exactly **one** changes outcome, and it is a **true positive**: the pipeline advanced past `monitor-merge` while PR #3277 was still unmerged — it merged **36 m 34 s later**. The 13 replays never occur because #1 stops the edge. The `monitor-deploy` one is exempt and unchanged, so **the Linear-Done backstop is provably untouched**.

So *"48% of advances are fabricated"* is **not** the blast radius. The gate is `fabricated ∧ phase-invertible ∧ probe-says-NOT-DONE` — 1/31 on the only sample that exists. That conjunction is why round 1's unconditional inversion was correctly blocked: it gated on the first conjunct alone.

Both cases are replayed as tests, including the discriminator that matters: the **same phase and ticket** with a throwing probe yields UNKNOWN, not a failure.

## Tests — 21

Covering: no UNKNOWN reason on any phase can produce a failed write (cross-product); only explicit NOT_DONE in enforce mode does; a non-boolean probe result is UNKNOWN rather than coerced (`undefined` is falsy — the arity-bug shape that produced 11 false negatives in CTL-1801); an exempt phase **never invokes the probe at all** (the daemon tick has a documented event-loop-stall history, so an exempt phase must not pay for a spawn).

## Next

- **PR-3** — wire in shadow at `flipSignalDoneOnSuccess`, which is already `async` at its call site, so an awaited probe blocks nothing. Threading `phase`/`ticket` through is the bulk of that change.
- **`enforce` is BLOCKED** on **CTL-1805** — the `monitor-merge → monitor-deploy` edge re-fired **13× in 55 s** from an unchanged signal file, and the mechanism is not yet established. That is 13 chances to write a terminal `failed` on a merged PR.

Full design, with a verification ledger stating how each claim was checked: `thoughts/shared/research/2026-08-12-ctl-1790-round4-design.md`.

### DIFF
--- a/plugins/dev/scripts/execution-core/outcome-verdict.mjs
+++ b/plugins/dev/scripts/execution-core/outcome-verdict.mjs
@@ -1,0 +1,215 @@
+// outcome-verdict.mjs — the three-valued outcome verdict behind CTL-1790.
+//
+// THE PROBLEM THIS EXISTS TO SOLVE. When a phase worker exits cleanly WITHOUT
+// declaring an outcome, the pipeline fabricates a terminal `done` from the clean
+// process exit (`sdk-success-flip`). Measured live on host mini after CTL-1789
+// shipped: of 31 phase advancements, 16 were `declared` and **15 were
+// `fabricated`**. The pipeline was inferring success roughly half the time.
+//
+// Three prior attempts to simply invert that — treat an undeclared exit as
+// `failed` — were each blocked by a real defect. All three died on the SAME
+// thing, and it is the reason this module exists:
+//
+//   **A boolean probe conflates "I checked, the work is absent" with "I could
+//   not check."** Every probe in work-done-probes.mjs returns `false` for both.
+//   Inverting on `false` therefore manufactures terminal failures out of GitHub
+//   outages, unresolvable worktrees, absent probes, and circular probes.
+//
+// So the verdict is THREE-valued, and the whole safety property is one rule:
+//
+//   **UNKNOWN NEVER WRITES `failed`.** On UNKNOWN the caller reproduces today's
+//   behavior byte-for-byte (the flip still writes `done`) and emits telemetry.
+//   An inversion may only fire on a probe that is decisive AND non-circular AND
+//   on a phase the allowlist admits.
+//
+// This mirrors the house rule already stated verbatim at recovery.mjs:3027-3028
+// — "STRICTLY gated on snap.isFresh — a stale/cold snapshot skips the
+// cross-check and suppresses exactly as before."
+//
+// Zero imports beyond the probe registry, so a bare-Node caller can load it.
+
+import { hasProbe } from "./work-done-probes.mjs";
+
+export const VERDICT = Object.freeze({
+  DONE: "done",
+  NOT_DONE: "not-done",
+  UNKNOWN: "unknown",
+});
+
+// Why a verdict came back UNKNOWN. Every one of these is a case where an
+// absence is NOT evidence, and the caller must fall back to today's behavior.
+export const UNKNOWN_REASONS = Object.freeze({
+  NO_PROBE_FOR_PHASE: "no-probe-for-phase", // hasProbe(phase) === false — teardown
+  PHASE_EXEMPT: "phase-exempt", // phase not in INVERTIBLE_PHASES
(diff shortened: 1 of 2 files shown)


ryanrozich:
Closing — superseded by the goal-loop pivot, not by a defect in this code.

This PR builds `outcome-verdict.mjs`: a three-valued DONE/NOT_DONE/UNKNOWN verdict so a probe could contradict the SDK success flip without manufacturing failures out of GitHub outages, missing worktrees and probe-less phases. It solves *make inference safe*.

Under goal-loops that problem ceases to exist. A worker declares its own outcome or nothing is recorded — there is nothing left to infer. CTL-1790 is rewritten accordingly: delete the fabricated `done`, write a terminal `abandoned`, emit `phase.<name>.abandoned.<ticket>`. No probe, no verdict, no flag, no shadow window.

Nothing here is being discarded for being wrong:
- **PR-1 (`e785a6e4`, the `matchesTicket` boundary fix) stands** — a live false-done vector, independent of the inversion.
- The round-4 design doc stays as history (`thoughts/shared/research/2026-08-12-ctl-1790-round4-design.md`). Its §0 verification ledger is still good; its §2 is obsolete.
- The un-timed `spawnSync` seams in `defaultRunGit`/`defaultRunGh` — the PR-1 companion that did not ship — are a real defect that survives the pivot and are being filed separately.

Branch `ctl-1790-outcome-verdict` left in place; the diff is two files and recoverable from this PR if the fork ever reverses.

PR #3341, opened 2026-08-13
fix(dev): CTL-1818 — count the events the forwarder throws away, on a surface that survives the throwing-away

ryanrozich:
Closes CTL-1818. Second ticket of **M1 — the event substrate is trustworthy** (sibling of #3325).

## The defect

`otel-forward` discards batches — `emitDrop`, reason `aged` or `terminal_4xx` — and **nothing measures it**. The two sibling signals are each covered by another instrument; this one is genuinely invisible:

| signal            | normally caught by                             | catches an aged drop?                                                    |
| ----------------- | ---------------------------------------------- | ------------------------------------------------------------------------ |
| `forward_failed`  | DLQ grows → CTL-1502 watchdog P1 (DLQ size)    | **No** — an aged record never rides the DLQ, so the file stays flat       |
| `forward_lag`     | checkpoint `lastForwardedTs` freezes → P2      | **No** — the checkpoint advances; a discard *is* read progress            |
| `forward_dropped` | an event on the log…                           | **No** — that event ships through the path that just discarded something  |

And two guards at the top of `emitDrop` (`!eventLogPath`, `isSelfBatch`) mean some discards emit **no event at all**, so they have no surface whatsoever.

## Measurement

Structured parse of the fleet's mirrored `~/catalyst/events/2026-08.jsonl` — name read as `attributes["event.name"] ?? event ?? name`, `grep` used only as a superset pre-filter and the decision made on the parsed name (8 substring-only lines were correctly rejected):

| scope                                     | drop events | discarded **records** |
| ----------------------------------------- | ----------: | --------------------: |
| fleet (mirror coverage 2026-08-08 → 08-13) | **87,706**  | **7,623,269**         |
| └ `mini-2`                                | 68,131      | 3,260,951             |
| └ `mini`                                  | 19,575      | 4,362,318             |
| by reason                                 | `aged` 87,706 · `terminal_4xx` **0** | |

Every health surface was green throughout. *Positive control for the instrument:* the same pre-filter over `recovery.tick` returns 557,150, and the parse returns non-zero for the name under test, so a zero here would have been a finding rather than a broken reader.

**Consumer search** (the negative claim): the only non-test occurrence of `catalyst.observability.forward_dropped` anywhere in this repo is the emitter itself, `otel-forward/lib/destinations/otlp.ts:245`. *Positive control for that search:* the same instrument over `reap-requested` returns the producer registry **and** its reaper consumer. In `catalyst-otel`, `provisioning/alerting/read-path-alerts.yaml` alerts on `forward_failed` — a **different name** — so `forward_dropped` has no external consumer either.

## The mechanism

`recordDrop` (`otel-forward/lib/drop-surface.ts`) is called from `emitDrop` **before** both early-return guards, so a discard is counted even when no event is written. Three sinks, none of which ride the pipe being reported on:

1. **Counters** — process-exact, per reason, in events (batches) **and** records.
2. **Marker** — `~/catalyst/otel-forward-drops.json`, atomic tmp+rename, `cumulative` totals carried across restarts, beside the checkpoint and the DLQ where an operator already looks:
   ```bash
   jq '{host, cumulative, window, alert}' ~/catalyst/otel-forward-drops.json
   ```
3. **Alarm** — the daemon's pino `.log`, which Alloy ships to Loki **independently** of this daemon's own OTLP egress.

There is deliberately **no event sink**: an alarm sourced from the forwarded stream measures the survivors and reads clean during exactly the outage it exists to detect.

The sustained-rate alert is a pure `classifyDropWindow` over a **bounded 60-bucket ring** (no unbounded sample list in a long-lived daemon): breach at `windowRecords >= thresholdRecords`, alert once the breach persists `>= sustainMs`. Raise and clear are **edge-triggered**, so a storm logs one `ERROR` rather than one per batch, and the forwarder's existing 30 s lag timer re-evaluates so a host that recovered and went quiet does not latch raised forever. Defaults (5 min / 1,000 records / 10 min) sit between the two regimes actually present in the measurement — a quiet day ≈ 813 records per 5-min window, the 08-10/08-11 storms ≈ 6,900 — and resolve **env > `forwarders.otlp.dropSurface` > frozen default**. A malformed override is ignored rather than silently disabling the counter.

**One mechanism, not two:** CTL-1817's count-every-occurrence/log-sparsely gate moved verbatim to `lib/sparse-warn.ts` as a factory. The tailer's unknown-shape alarm and this drop alarm each build their own gate, so a flood of one cannot exhaust the other's key budget.

## Alert-only, on purpose

This adds **no** predicate to `classifyDaemonStuck`. That function's `tripped` list is wired straight to `restart()` in the CTL-1502 probe (`daemon-watchdog-probe.mjs`), so a third member would **restart otel-forward** on a sustained aged drop. Restarting does not fix an aged drop — it is a Loki-accept-window / backlog condition — and would only lose the in-memory buffer. The ticket's AC asks that an alert fire; it never authorizes a restart. **No watchdog file is touched.**

## Verification

| suite                                                  | this branch                 | merge base (`cf37127`)      |
| ------------------------------------------------------ | --------------------------- | --------------------------- |
| `cd plugins/dev/scripts/otel-forward && bun test`      | **236 pass / 0 fail**, 18 files | 211 pass / 0 fail, 16 files |
| `bun run typecheck` (`tsc --noEmit`)                   | clean                       | clean                       |

The delta is exactly the +25 added here; no pre-existing red imported. Baseline measured by restoring the merge-base tree into this worktree, not by inference.

**Positive control** (required by the ticket's own AC, and run for real): reverting the single wiring line in `emitDrop` takes `otlp-drop-surface.test.ts` from 6 pass → **2 pass / 4 fail**. The 2 that stay green are the two that *should* be invariant — the delivered-batch case and the bypassed-recorder case. The line was then restored and the suite re-run green. The AC's "test-only bypass" is the injectable `dropRecorder` seam (default = the real recorder, so the wiring cannot be forgotten at a call site); no production kill-switch was added.

Also covered — the three cases no event-based consumer can satisfy: a discard with **no event log configured**, a **self-batch** discard whose event is loop-guarded away, and a discard while the **OTLP endpoint is unreachable**.

`lib/destinations/otlp.test.ts` gains a pinned `CATALYST_DIR` (re-asserted per test, since bun runs every file in one process). Found the hard way: the first full-suite run after wiring wrote a bogus marker into the real `~/catalyst/otel-forward-drops.json` on this machine. That file was removed and the suite now cannot reach it.

## Stale premises in the ticket, left uncorrected on purpose

1. **The catalog edit the ticket asks for must NOT be made.** CTL-1818 says `2026-08-13-event-automation-catalog.md` §5(a) "recommends deleting all three families … update that cut-list row so the delete is not actioned". That row has **already been corrected** and now reads **"⛔ REJECTED — do not delete"** for all three, because `forward_failed` has a **live external consumer**: `catalyst-otel/provisioning/alerting/read-path-alerts.yaml:252-276`, the only "the stack goes blind" guard in the stack (verified by reading both files). Making the edit as written would *reverse* that correction and greenlight deleting an event a provisioned Grafana alert reads. The rows are left intact. (The catalog lives in `thoughts/`, a separate repo, and is untouched by this PR either way.)

2. **The volume numbers do not disagree — they are differently scoped, and neither is a month.** The ticket's "19,569 a month (mini)" is the **`mini`-only** subset; the catalog's 87,216 is the **fleet** total of the same mirrored file. My measurement reproduces both (19,575 and 87,706, a day later). But the laptop's mirror only covers **2026-08-08 → 08-13**, so both figures are ~6 days of mirrored log presented as "/month" — the real monthly rate is several times higher. Neither figure counts **records**, which is where the loss actually lives: 7.6 M.

## Deferred (not in this PR)

- **A Grafana rule** over the `otel-forward.log` stream (`| msg =~ "SUSTAINED EVENT LOSS"`, or the `drop_records_total` field) belongs in the sister repo `catalyst-otel`; a malformed provisioned rule file crash-loops the **shared** Grafana, so it is described rather than written here.
- **A HUD / orch-monitor renderer** over `~/catalyst/otel-forward-drops.json` — the same follow-up the CTL-1502 watchdog marker is still waiting on (its Codex P2).
- **Root-causing the aged drops themselves.** This PR makes the loss visible; it does not reduce it. 7.6 M records in 6 days at 100% `aged` says records are reaching the sender well past `lokiAcceptWindowMs` — worth its own ticket now that there is a number to watch.
- **PostHog / Cloudflare AE** discards. The surface is destination-agnostic by construction (it records a reason and a count), but only the OTLP sender is wired here.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -630,6 +630,42 @@ mode, restarts the stuck daemon exactly once per breach episode.
   not just `kill -0`, and fail closed — a recycled pid is treated as a stale file, never a kill
   target. Covered by `__tests__/daemon-watchdog-supervision.test.sh`.
 
+### Forward-drop surface (CTL-1818)
+
+The watchdog above catches a **wedged** forwarder. A forwarder that is working perfectly and
+**discarding** events is a different failure, and none of the three existing instruments can see it:
+`emitDrop` (`otel-forward/lib/destinations/otlp.ts`) fires for `aged` or `terminal_4xx`, and an aged
+record **never rides the DLQ in the primary path** — so the watchdog's DLQ-size predicate stays flat,
+the checkpoint keeps advancing (a discard *is* read progress), and the
+`catalyst.observability.forward_dropped` event it writes ships through the very path that just
+discarded something. Measured on the fleet's mirrored `2026-08.jsonl` (coverage 08-08 → 08-13):
+**87,706 drop events carrying 7,623,269 discarded records, 100% `reason: "aged"`** — read by nothing
+in this repo, and by nothing in `catalyst-otel` either (the provisioned Grafana rule reads
+`forward_failed`, a different name).
+
+- **Where it is counted** — `recordDrop` (`otel-forward/lib/drop-surface.ts`) is called from
+  `emitDrop` **before** its `!eventLogPath` / `isSelfBatch` guards, both of which suppress the event
+  entirely. Counters are process-exact, per reason, in events (batches) **and** records.
+- **Three sinks, none of which ride the failing pipe** — (1) the exact in-memory counters;
+  (2) a durable marker `~/catalyst/otel-forward-drops.json` (atomic tmp+rename, `cumulative` totals
+  carried across restarts, ≤1 s write throttle bypassed on any alert transition); (3) the daemon's
+  pino `.log`, Alloy-shipped independently of its own OTLP egress. Deliberately **no** event sink —
+  an alarm sourced from the forwarded stream measures the survivors and reads clean during exactly
+  the outage it exists to detect.
+- **Sustained-rate alert** — pure `classifyDropWindow` over a bounded 60-bucket ring: breach when
+  window records `>= thresholdRecords`, alert when the breach persists `>= sustainMs` (defaults 5 min
+  / 1,000 records / 10 min, resolved env > `forwarders.otlp.dropSurface` > frozen default). Raise and
+  clear are **edge-triggered**, so a storm logs one `ERROR`, not one per batch; the forwarder's
+  existing 30 s lag timer re-evaluates so a host that recovered and went quiet does not latch raised
+  forever.
+- **ALERT-ONLY, by ruling.** This is *not* a third member of `classifyDaemonStuck`'s `tripped` list —
+  that list is wired straight to `restart()` in the CTL-1502 probe, and restarting a forwarder does
+  not fix a Loki-accept-window/backlog condition; it only loses the in-memory buffer. A Grafana rule
+  over the `.log` stream is the follow-up, and belongs in the `catalyst-otel` repo.
+- **One mechanism, two callers** — the CTL-1817 count-every/log-sparsely gate moved to
+  `otel-forward/lib/sparse-warn.ts` (semantics unchanged); the tailer's unknown-shape alarm and this
+  drop alarm each build their own gate so a flood of one cannot exhaust the other's key budget.
+
 ### Two-axis worker state & the recordTransition chokepoint (CTL-764)
 
 Every worker ticket has **two orthogonal axes** — never blurred:
--- a/plugins/dev/references/event-forwarding.md
+++ b/plugins/dev/references/event-forwarding.md
@@ -88,6 +88,44 @@ not duplicated here.
 with attribute `catalyst.observability.drop_reason` (`"aged"` or `"terminal_4xx"`) and
 `body.payload.count`. Loop-guarded by `isSelfBatch` (same as `forward_failed`).
 
+### Drop surface — `~/catalyst/otel-forward-drops.json` (CTL-1818)
+
+A discard is the one forwarder failure **no other instrument can see**:
+
+| signal          | how it is normally caught                                      | does it catch an aged drop?                                            |
+| --------------- | -------------------------------------------------------------- | ---------------------------------------------------------------------- |
+| `forward_failed` | DLQ grows → CTL-1502 watchdog P1 (DLQ file size)               | **No** — an aged record never rides the DLQ, so the file stays flat     |
+| `forward_lag`    | checkpoint `lastForwardedTs` freezes → watchdog P2             | **No** — the checkpoint advances; a discard *is* read progress          |
+| `forward_dropped`| an event on the log…                                           | **No** — that event ships through the path that just discarded something |
+
+So the drop is accounted for **on the host**, out of band of the forwarder's own OTLP egress:
+
+- **Counters** — exact, per reason, in both events (batches) and records. Recorded in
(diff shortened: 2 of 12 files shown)


ryanrozich:
## Remediation: the two drop-surface wiring lines are now load-bearing

An adversarial verify pass found that both production wiring lines in `otel-forward/index.ts` could be **deleted with the whole suite still green** (236 pass / 0 fail either way) — the "tested the pure function, not the gate in front of it" defect. The unit suites call `resolveDropSurfaceConfig` and `evaluateDropSurface` directly, so they passed whether or not the daemon ever called them.

| line | what its deletion silently costs |
| --- | --- |
| `configureDropSurface(cfg.otlp.dropSurface)` | the documented `forwarders.otlp.dropSurface` layer goes **dead** — the surface runs on env + frozen defaults forever, while `configuration.md` (added in this PR) declares that layer authoritative |
| `evaluateDropSurface()` on the 30 s lag tick | a recovered host latches `alert.raised: true` **forever** — `drop-surface.ts`'s clear branch is reachable from this tick and nothing else |

New `plugins/dev/scripts/otel-forward/index-drop-wiring.test.ts` asserts both **behaviourally**, from a spawned probe. A subprocess because both are module-scope facts (`configureDropSurface` runs on `import`, and `bun test` shares one module registry across files, so an in-process import would be silently order-dependent). The probe pins `CATALYST_DIR` / `CATALYST_EVENTS_DIR` / `CATALYST_CONFIG_PATH` so nothing lands in the real `~/catalyst`, and builds its env from scratch so a `CATALYST_FORWARD_DROP_*` in the shell cannot stand in for the config tier under test.

1. Import `index.ts` with a config file carrying `111000 / 222 / 333000` — all three unlike the frozen defaults — and read them back off `dropSurfaceSnapshot().config`.
2. Latch a sustained-loss alert on an injected clock, call `emitLag()` once, and watch `alertRaised` **and** the marker's `alert.raised` go false. The BEFORE assertions are the positive control: they prove the alert was genuinely raised, so the clear is a real clear rather than a never-raised vacuous pass.

`emitLag` is exported for (2); the `lagTimer` remains its only runtime caller. No existing test was weakened.

### Mutation control — full `bun test` in `plugins/dev/scripts/otel-forward` (239 tests)

| mutation | result | which probe fails |
| --- | --- | --- |
| _baseline (fixed)_ | **239 pass / 0 fail** | — |
| delete `configureDropSurface(...)` | 238 pass / **1 fail** | config probe |
| delete `evaluateDropSurface()` | 238 pass / **1 fail** | tick probe |
| `configureDropSurface({})` — plausible wrong-arg refactor | 238 pass / **1 fail** | config probe |
| move `evaluateDropSurface()` below the cold-start skip | 238 pass / **1 fail** | tick probe |

Each mutation fails exactly its own probe and leaves the other green, so the two assertions are independent rather than one coupled check. `tsc --noEmit` clean; suite stable across three consecutive runs.

PR #3353, opened 2026-08-14
fix(dev): CTL-1835 — close the four residual gaps from #3349, and prove each closure

ryanrozich:
Closes CTL-1835. Follow-up to CTL-1831 (#3349, merged as `3f60190c6`): the four residual gaps that PR's round-4 adversarial pass found and deliberately did not chase, so they were recorded rather than lost.

The functional change is **two lines**. Everything else is the coverage that proves them.

### 1. One unguarded mutation — `wrongCopy` dropped from the `mismatched` verdict

Round 4 ran 17 mutations; 16 went RED, **M7 did not**. Its siblings *are* guarded (M6 on `matched` → RED, M8 on the `inconclusive` reason → RED), so the field was covered on two of its three verdicts and blind on the third.

Proven closed with a **negative control**, which is the only way to show the new test is what closes it:

| M7 — delete `lock-resolution-audit.mjs:715` | result |
| --- | --- |
| against `origin/main`'s test file | **52 pass / 0 fail — GREEN** (the gap was real) |
| against this branch's test file | **56 / 1 — RED** |

### 2. A stale test-count citation

Corrected `175` → **196 (139 + 57)**, removed a stale `97`, and `7 I13` → **11**; the `38` reproduced and was kept. Each now names its instrument, bun version and host. The head commit measured **189**, not 175.

### 3. `plugin-refresh.mjs:437` — a second glob guard was mutation-uncovered

Pre-existing (CTL-1628), explicitly out of #3349's scope so it would not be mistaken for coverage.

| drop `\|\| entry.includes("*")` at `:437` | result |
| --- | --- |
| against `origin/main`'s test file | **137 pass / 0 fail — GREEN** (reproduces the documented uncovered state exactly) |
| against this branch's test file | **137 / 2 — RED** (both new tests, from two independent directions) |

**No source change to `plugin-refresh.mjs`** — the fix is entirely new tests, so this PR does not touch the region CTL-1834 is concurrently editing.

### 4. `grep` silently returned nothing on `lock-resolution-audit.mjs` — root-caused, not documented around

The ticket asked to either fix the cause or add a warning comment. The cause turned out to be **one byte**.

`file` classified the source as `data`, so **plain `grep` returned zero without saying why** — a silent-zero instrument sitting inside the module that is the fleet's only defence against a silent stale install. It had already cost one agent a false negative mid-analysis.

Cause: **exactly one literal NUL byte at offset 23048, line 444**, an intentional memo-key separator written as a raw byte instead of an escape:

```js
const cacheKey = `${fromDir}<NUL>${id}`;
```

Fixed by hoisting it to `export const MEMO_KEY_SEP = "\0"`. The string value is byte-identical; only the *source encoding* changes.

| check | before | after |
| --- | --- | --- |
| `file` | `data` | **`Unicode text, UTF-8 text`** |
| `grep -c 'wrongCopy'` | *(silent zero)* | **9** |
| `grep -ac 'wrongCopy'` | 9 | 9 — now agree |

The separator's *identity* is asserted, not just its behaviour, because a later "cleanup" to a visible character would silently reintroduce a collision. With `/` as the separator, `(fromDir="a", id="b/c")` and `(fromDir="a/b", id="c")` produce the **same** key and the memo returns the wrong cached resolution. Verified both directions: distinct under `\0`, equal under `/`.

## Verification

40 mutations run, **40 RED, 0 GREEN** — every added production line, all 23 new assertions (each expected value perturbed so it *must* fail), and 8 fixture-erosion probes.

The harness was proven capable of RED before any green was trusted, because **round 4 of the parent PR caught and discarded a broken mutation harness** that silently mis-parsed ANSI and reported every mutation GREEN. This one strips ANSI, fails closed on unparseable output, and checks file md5s per run.

## Notes

- **Gap 4's hazard bit while fixing gap 4**: writing the tests put three literal NUL bytes into the test file — they paste invisibly and nothing complains. Caught by byte-scanning, not by reading. The NUL guard now covers the test file too, since that is demonstrably where the byte comes back.
- **Two instruments returned false zeros during this work**, both caught by positive control: `git grep -E '\b137\b'` returned 0 while plain `grep` returned 41 (POSIX ERE has no `\b`), and a `cd`-broken `git show` produced empty files that prettier then cheerfully passed. Both are exactly the defect class this ticket is about.
- **Declined, with reason**: two other citations in `execution-core-tests.yml` (`273`/`120` "on this branch") do not reproduce today, but they record a historical *branch-vs-merge-base delta* justifying why those steps exist. Overwriting them with today's values would destroy their meaning, and that file is a known merge-conflict hot spot with CTL-1834 in flight. Left as-is deliberately.
- **One pre-existing broker-suite failure** (`held-label literals stay in lock-step with the board`) is unrelated: reproduced at HEAD with these changes stashed (1004 pass / 1 fail at HEAD vs 1011 / 1 fail here — the delta is exactly the 7 tests added).
- No new test files, so the hand-enumerated CI list needed no additions; both files were already registered. YAML re-validated as parsing.

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -526,24 +526,35 @@ jobs:
 
       - name: plugin-refresh + lock-resolution-audit tests (CTL-1831)
         # `grep plugin-refresh .github/workflows/` returned only two COMMENTS
-        # before this step — plugin-refresh.test.mjs (97 tests guarding the
-        # merge-to-main checkout refresh, the deps install, the stale-index.lock
-        # clear and the pull-owner cutover) had never run in CI at all. Positive
-        # control: the same grep for `worker-state-projection` returned the `run:`
-        # line directly above. CTL-1831 adds the install-must-relink guard to that
+        # before this step — plugin-refresh.test.mjs (guarding the merge-to-main
+        # checkout refresh, the deps install, the stale-index.lock clear and the
+        # pull-owner cutover) had never run in CI at all. Positive control: the
+        # same grep for `worker-state-projection` returned the `run:` line
+        # directly above. CTL-1831 adds the install-must-relink guard to that
         # same file plus its own audit suite, and an unlisted suite is a guard that
         # lands with no gate — this job's list is hand-enumerated, not globbed.
         #
-        # Both files run: measured 175 tests / 0 fail on this branch, so there is
-        # no pre-existing red to import. (Instrument: `bun test
-        # plugin-refresh.test.mjs lock-resolution-audit.test.mjs` from this step's
-        # working-directory, bun 1.3.5, macOS arm64, single run — the count is
-        # deterministic, no spread. On a checkout WITHOUT the broker package's
-        # pino installed, one of the 175 fails to load rather than failing an
-        # assertion; CI installs it, so CI sees 175/0.) They cannot join the
-        # `bun test` allowlist below — that step's working-directory is
-        # execution-core, and plugin-refresh.test.mjs reaches router.mjs, which
-        # needs the BROKER package's own pino.
+        # Both files run: measured 196 tests / 0 fail (139 plugin-refresh + 57
+        # lock-resolution-audit), so there is no pre-existing red to import.
+        # INSTRUMENT: `bun test plugin-refresh.test.mjs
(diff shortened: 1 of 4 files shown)


ryanrozich:
@codex review

PR #3359, opened 2026-08-14
fix(meta): manifest the 17 bash attribute keys that rotted main red

ryanrozich:
**This unblocks every open PR in the repo.** `otel-attribute-audit — emitter=sh (canonical-event.sh) > no_missing_audit_entry` fails on **clean `origin/main`** (`890c12d91`), so it fails on every branch regardless of content.

## Why nobody saw it

`quality` is a **PR-only** check. Measured on `check-runs` for four consecutive main commits:

| commit | `quality` |
| --- | --- |
| `890c12d91` chore: release main | **NOT RUN** |
| `725922a8e` CTL-1835 | **NOT RUN** |
| `3f60190c6` CTL-1831 | **NOT RUN** |
| `388edf41b` CTL-1809 | **NOT RUN** |

So main can go red with no red run on main — the same silent-rot class as the release-please path skip. Filed separately as **CTL-1840** so the class gets closed, not just this instance.

## The fix

`extractShellAttributeKeys` deliberately captures **every** `"key":` in `canonical-event.sh` minus TS-interface-owned keys — its docstring says the remainder "is what the bash emitter contributes beyond the TS schema." So the manifest is meant to document all of them. Two merged changes added keys without entries:

- **CTL-1795's flat→OTel promotion map** — the six *left-side* v1 field names (`ticket`, `phase`, `bg_job_id`, `branch`, `orch_id`, `dominant_phase`). Classified **`conforming`**, deliberately not `rename-to`: nothing emits them under those names, so claiming a pending rename would be untrue *and* would require a `remediationCluster` for work that already shipped. Each note records the target it is promoted to.
- **CTL-1809's oversized-line tombstone** — it builds a complete v2 record inline, so the extractor also sees eight OTLP **log-record structural fields** (`ts`, `observedTs`, `severityText`, `severityNumber`, `body`, `message`, `attributes`, `resource` — conforming by construction) plus the tombstone's own three `catalyst.event.oversized.*` attributes (legitimately-custom).

`docs/otel-attribute-audit.md` regenerated with `bun run audit:gen` (the Phase-3 staleness gate).

## The gate is repaired, not silenced

The obvious failure mode for a fix like this is to make the check pass by widening it until it can never fail. Proven otherwise — inserting an unmanifested key into `canonical-event.sh`:

```
probe key inserted: 1
✗ otel-attribute-audit — Phase 2 > emitter=sh > no_missing_audit_entry
 22 pass / 1 fail
```

Restored: **23 pass / 1 skip / 0 fail**.

(First attempt at that proof inserted **0** keys because of a bad `perl` guard, and reported a meaningless GREEN — caught by asserting the probe count before trusting the result.)

### DIFF
--- a/plugins/dev/scripts/orch-monitor/docs/otel-attribute-audit.md
+++ b/plugins/dev/scripts/orch-monitor/docs/otel-attribute-audit.md
@@ -10,10 +10,10 @@ Do **not** edit this file directly.
 
 | Classification | Count |
 | --- | --- |
-| ✓ Conforming | 20 |
+| ✓ Conforming | 34 |
 | → Rename-to | 37 |
-| ● Legitimately Custom | 40 |
-| **Total** | 97 |
+| ● Legitimately Custom | 43 |
+| **Total** | 114 |
 
 ## Classification by Emitter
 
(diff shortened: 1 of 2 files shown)


ryanrozich:
Superseded by #3367, merged as `b8cb40b5f`, which carries this manifest commit verbatim alongside the governance-journey fix.

The two PRs were deadlocked on each other: #3359's quality job failed on `governance-journey.contract.test.ts` (a filesystem-dependent `readdirSync` tie-break, CTL-1723) and #3367's failed on the `otel-attribute-audit` this PR fixes. Each was red only because of the other, so one combined green run was the only way out.

Verified in main rather than assumed: 6 `CTL-1795 flat→OTel map` entries and 3 `catalyst.event.oversized.*` keys are present at `b8cb40b5f`.

One correction preserved for the record: the key count is **17**, not 19. Measured with the audit's own failure output on `3f60190c6`, `890c12d91`, and the fix branch `c94e2e71d` — all 17, with an empty diff between the sets in every direction — and confirmed on Linux CI, where this PR's audit failure disappeared entirely. The merge commit message reflects 17.

CTL-1840 remains open as the class: `quality` is PR-only and path-filtered, so nothing prevents a red from landing and nothing on main re-checks.

PR #3432, opened 2026-08-16
fix(dev): pin @catalyst-cloud/sdk 0.8.5, ending the pr_* frame loss

ryanrozich:
## The data loss this ends

The mirror has been emitting change frames for four entities **no published schema defined** — `pr_commits`, `pr_review_comments`, `pr_conversation_comments`, `pr_events` — so every client dropped them regardless of version.

They are **dropped, not retried**: the cursor advances straight past a failed frame. That makes it *permanent per-entity data loss underneath a cursor that looks perfectly healthy* — the failure mode no existing instrument was watching.

Measured on the fleet, 20k-line log sample:

```
414  pr_commits
321  pr_review_comments
166  pr_conversation_comments
---
901  failed frames
```

**Positive control** in the same sample: 3,926 frames applied, and `issues` failures = **0**. So the zero above is evidence, not a broken query.

## The fix

`schema@0.1.12` defines all four — verified against the published tarball's `CREATE TABLE` statements, not the changelog. `0.8.5` pins `replicate@0.1.6` exactly, and `0.1.6` pins `schema@0.1.12` exactly, so both halves still name one schema and the CTL-1895 invariant holds across the bump.

## ⭐ Also bumps the ROOT schema pin, which the SDK bump does not move

The root `package.json` declared `@catalyst-cloud/schema: 0.1.10` **independently**, and `cloud-sync-schema-identity.mjs` imports that **bare** specifier while the replica applies with whatever the SDK resolves.

Bumping only the SDK would have left `bare = 0.1.10` against `sdk = 0.1.12` — precisely the **CTL-1869 split** (a host reporting a schema identity three migrations away from the one it applies), re-opened by the very upgrade meant to close a different version skew.

Caught by the one-copy gate on a fresh tree. It would **not** have appeared in any resolution check that starts from the SDK.

## Verification (off disk, following symlinks, from every importing position)

| resolution path | version |
|---|---|
| bare `@catalyst-cloud/schema` | 0.1.12 |
| schema **via sdk** | 0.1.12 |
| schema **via replicate** (the writer's `KNOWN_COLUMNS` source) | 0.1.12 |
| MIGRATED vs WRITTEN | **agree** |

The store retains an orphaned `schema@0.1.10` entry after the upgrade — bun's documented behaviour, and why a store-level copy count is only meaningful on a *truly fresh* tree. No resolution path reaches it; the authoritative check is what each importer resolves, and all three agree.

The CTL-1869 identity guard passes **17/17** here, exercising the unified-tree branch added in #3426 — a tree with one schema copy is a **result**, not an inconclusive test.

## Policy

Pin stays **exact**. A bump is a deliberate PR, never a range that floats — a caret is what let the running version drift from the declared one in the first place.

## Rollout

Same verified loop as #3426: merge → plugin-refresh delivers → restart mini-2 → mini → laptop, verifying LOADED per host and `pr_*` frames **applying** (0 failed) rather than trusting the version string.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "catalyst",
       "dependencies": {
-        "@catalyst-cloud/schema": "0.1.10",
+        "@catalyst-cloud/schema": "0.1.12",
         "@modelcontextprotocol/sdk": "1.29.0",
       },
       "devDependencies": {
@@ -34,7 +34,7 @@
       "name": "catalyst-execution-core",
       "dependencies": {
         "@catalyst-cloud/read-model": "^0.1.0",
-        "@catalyst-cloud/sdk": "0.8.4",
+        "@catalyst-cloud/sdk": "0.8.5",
         "@modelcontextprotocol/sdk": "1.29.0",
         "@opentelemetry/api": "^1.9.1",
         "@opentelemetry/exporter-trace-otlp-http": "^0.219.0",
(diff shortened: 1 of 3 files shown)


ryanrozich:
@codex review

PR #3436, opened 2026-08-16
fix(dev): CTL-1892 — the fleet has TWO app identities and both rotated; only one was seeded

ryanrozich:
Shipped gap in #3428, found by **enumerating** every actor named `Catalyst*` in the replica rather than reasoning from the one identity already known.

## The miss

The fleet runs an **orchestrator** app-actor *and* a **worker** app-actor. Both have rotated, on the same day, in the same way — Linear appends a collision suffix when an app re-mints onto a taken handle:

| retired | → | current | seeded in #3428? |
|---|---|---|---|
| `catalystorchestrator` (`f51bc697`) | → | `catalystorchestrator2` (`ba2989f1`) | ✅ |
| `catalyst` (`72824b58`) | → | `catalyst2` (`6dd38c1a`) | ❌ **missed** |

So the retired **worker** — **1,245 comments** — still read as a third party. Its echoes would dispatch: exactly the defect CTL-1892 closed, for the other identity.

## It meets the admission bar more cleanly than the pair already seeded

| | comments | window | overlap |
|---|---|---|---|
| `catalyst` | 1,245 | 2026-06-02 → **2026-08-10** | — |
| `catalyst2` | 97 | **2026-08-11** → today | **zero** |

The orchestrator pair both touched 08-10; this one has a clean gap. Same display name, successor handle carries the `2` suffix, named by no config live or backup. Measured with a **positive control** — the same query shape on the known-good orchestrator pair reproduces its documented handoff.

## ⚠️ The generalisable miss

I seeded from the identity the investigation had surfaced, instead of enumerating the population. One rotation was visible because a stray actor had already been noticed; the other was equally present and simply had not been looked for.

**The fix for "we found A" is to ask what the SET is, not to record A.**

## Test changes

Assertions now check **accumulation** rather than an exact identity set — an exact-equality assertion here just asserts the absence of the deliberate seeds, and broke on this change for that reason. Added: each seeded entry must name the successor it was superseded by (making each a checkable claim rather than an allowlist entry), and no **current** identity may appear in the retired list.

42 tests. Mutation-verified: removing the worker seed fails 3.

### DIFF
--- a/plugins/dev/scripts/execution-core/linear-bot-identity.mjs
+++ b/plugins/dev/scripts/execution-core/linear-bot-identity.mjs
@@ -82,6 +82,21 @@ export const KNOWN_RETIRED_IDENTITIES = Object.freeze([
       "name 'Catalyst Orchestrator' as its successor, whose handle carries Linear's collision " +
       "suffix '2'; same-day handoff with no overlap; named by no config, live or backup (CTL-1892)",
   },
+  {
+    // ⚠️ The fleet runs TWO app identities, not one — an ORCHESTRATOR and a WORKER — and
+    // each has rotated. The first cut of this list seeded only the orchestrator pair, so
+    // the retired WORKER's 1,245 comments still read as third-party: the same defect this
+    // module exists to close, for the other identity. Found by enumerating every actor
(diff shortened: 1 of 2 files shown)


ryanrozich:
@codex review

PR #3472, opened 2026-08-17
fix(dev): CTL-1921 — every new worktree inherited 1.3 GB of other agents' scratch checkouts

ryanrozich:
`create-worktree.sh` copied `.claude` wholesale — and `.claude/worktrees/` is where the agent harness parks its own `wf_*` / `agent-*` working trees, several with their own `node_modules`, some nested. So **every** worktree created inherited a byte-for-byte copy of **every** harness worktree that happened to exist at the time. COORD's disk sweep traced the 150 GB peak here and freed **14 → 156 GiB**.

Measured on the shared checkout the day this was filed:

```
.claude              1.3G
.claude/worktrees    1.3G   <- 13 entries
everything else      ~40K   <- rules, prompts, config*.json, settings.local
```

The part actually worth copying is **~0.003%** of what was being copied.

## Two independent changes

1. **Skip the machine-local scratch** — `worktrees/` and `debug/`. Done **entry-by-entry**, not copy-then-prune: pruning afterwards still writes the 1.3 GB first, which is the entire cost being removed.
2. **Clone instead of copy where the filesystem allows.** APFS `cp -c` and GNU `cp --reflink=auto` are both copy-on-write. The form is **probed once against a real temp file** rather than sniffed from `uname` — a Mac can have GNU coreutils first on PATH and a Linux box can be running busybox — and falls back to plain `cp -R`, which is always correct.

> The exclusion is what makes this correct; the clone only makes it fast. A clone still materialises metadata per file, and on a non-reflink filesystem it is a full copy again — so the exclusion must not be dropped on the grounds that cloning made it cheap. That reasoning is in the code, not just here.

**CTL-990's invariant is preserved**: tracked config is still restored to branch state after the copy, so a locally-dirty `.claude/config.json` in the main checkout does not dirty the new worktree.

## Verification — 9-case mutation matrix, 9/9 killed

New `create-worktree-scratch-exclusion.test.sh`, **18 assertions**. The headline is a **byte** assertion rather than a path-absence one:

```
PASS: copied .claude is 16K vs source 2072K — the scratch bytes were never written
```

`worktrees/ is not there` also passes if the copy silently stopped copying **anything**, so every exclusion case is paired with a positive control that the real config *did* arrive — including a **dotfile** (the glob enumerates those separately) and a config **subdirectory** (a non-recursive `cp` would drop it).

### ⛔ The matrix found a false-positive assertion in my own test

Deleting the `.catalyst` copy **entirely** SURVIVED. The only `.catalyst` assertion was on its *tracked* `config.json` — and the CTL-990 block runs `git checkout -- .catalyst` in the new worktree, which materialises every tracked file from the branch **whether or not the copy ran**. The assertion was being satisfied by something other than the thing under test.

Fixed with an **untracked** `.catalyst` file, which is the only kind that can tell those two apart.

### Two mutations recorded as equivalent rather than chased

Removing the unmatched-glob guard, and a probe returning an invalid flag. The per-entry `cp … 2>/dev/null || cp -R … || true` absorbs both, so neither has an observable effect — contorting a test to catch them would assert on implementation rather than behaviour. Recorded in the harness with that reasoning instead of being left as an unexplained survivor.

### Honest limit, stated in the test

The byte assertion measures the **destination** tree, so it does **not** catch a "fix" that copies everything and deletes afterwards. Detecting that needs peak-usage or timing measurement, which is flaky in CI; the guard against it is the per-entry loop plus the asserted skip line. (An earlier draft of that comment claimed the assertion covered it. It does not, and the comment now says so.)

## Pre-existing, unrelated

`create-worktree-base-ref.test.sh` fails one assertion (worktree HEAD vs base tip). Confirmed **pre-existing** by running that file against `origin/main`'s `create-worktree.sh` in this same worktree: **21 passed / 1 failed both ways**, and the assertion concerns branch selection, not config copying.

Refs CTL-1921.

### DIFF
--- a/plugins/dev/scripts/__tests__/create-worktree-scratch-exclusion.test.sh
+++ b/plugins/dev/scripts/__tests__/create-worktree-scratch-exclusion.test.sh
@@ -1,0 +1,214 @@
+#!/usr/bin/env bash
+# CTL-1921: create-worktree.sh must NOT copy the agent harness's own scratch
+# checkouts into every new worktree.
+#
+# `cp -R .claude` copied the whole directory, and `.claude/worktrees/` is where the
+# harness parks `wf_*` / `agent-*` working trees — several with their own
+# `node_modules`. Measured on the shared checkout the day this was filed:
+#
+#   .claude              1.3G
+#   .claude/worktrees    1.3G   <- 13 entries
+#   everything else      ~40K
+#
+# So each new worktree inherited a byte-for-byte copy of ~1.3 GB of other agents'
+# scratch to obtain ~40 KB of config. The disk sweep that found it freed
+# 14 -> 156 GiB.
+#
+# ⛔ THE LOAD-BEARING ASSERTION IS THE ONE THAT MEASURES BYTES, NOT THE ONE THAT
+# CHECKS A PATH IS ABSENT. "worktrees/ is not there" also passes if the copy
+# silently stopped copying anything at all, so every exclusion case is paired with
+# a positive control proving the real config DID arrive.
+#
+# Run: bash plugins/dev/scripts/__tests__/create-worktree-scratch-exclusion.test.sh
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../../.." && pwd)"
+CREATE_WT="${REPO_ROOT}/plugins/dev/scripts/create-worktree.sh"
+
+FAILURES=0
+PASSES=0
+fail() {
+	FAILURES=$((FAILURES + 1))
+	echo "  FAIL: $1"
+}
+pass() {
+	PASSES=$((PASSES + 1))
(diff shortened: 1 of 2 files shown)


ryanrozich:
## Verified against the real 1.3 GB `.claude`, not just the scratch fixture

The test harness plants a 2 MiB fixture. This is the patched copy helper run against the actual shared checkout that produced the 150 GB peak:

```
probe flag: [-Rc]
   ↳ skipping .claude/worktrees (machine-local scratch, CTL-1921)
elapsed_sec=0

source .claude: 1.3G
copied .claude:  60K
entries copied:  10
```

**1.3G → 60K, in under a second.** The probe selected `-Rc` (APFS clone) on this host.

Every real config entry survived, including both dotfiles — which matters because the dotfile glob is enumerated separately and a naive `for entry in "$src"/*` would silently drop them:

```
.gitignore  .personal  config.example.json  config.json  config.template.json
plugins  prompts  rules  scheduled_tasks.lock  settings.local.json
```

Run read-only into a temp dir — no worktree created, nothing written to the shared checkout.

PR #3535, opened 2026-08-18
feat(dev): CTL-1929 — the GitHub leg's enforce flip, suppressing smee PER NAME

ryanrozich:
## What this is

`enforce` on the GitHub leg was refused **entirely**, and the refusal was correct but too coarse. This producer cannot emit `github.pr.merged` (**CTC-691**, no `merge_commit_sha`) or `github.check_suite.completed` (**CTC-667 item 4**, the mirror stores no suite row), and suppressing smee for a name with no replacement takes out the CI wait and the whole merge→deploy chain fleet-wide with nothing behind them. So `enforce` degraded to shadow and **nine fully-covered names sat behind two uncovered ones**.

This makes enforce reachable at the granularity the problem actually has: **the gate suppresses smee per NAME.** 9 of the 12 names the router consumes cut over; the 3 gaps hold back only themselves and each names its own ticket in the capture record.

⛔ **No host changes.** `CATALYST_GITHUB_FEED` is `off` everywhere except mini-2 (`shadow`). Merging this flips nothing — `createGithubFeedGate` returns `null` for `off`, so the broker's tail skips the block entirely and routing is byte-identical to pre-CTL-1929.

## ⛔ The finding: `github.push` is excluded, and that corrects a claim this feature shipped

`github-feed-source.mjs`'s header and `packages/schema/src/mirror.ts`'s comment on `pushes` both justify the `(repo_id, ref)` collapse the same way:

> *"Rebase detection asks 'has this ref moved since I last looked', which a per-ref head answers exactly."*

**That describes a polling consumer. We do not have one.** Measured against the code:

- `broker/router.mjs:1582` — on `github.push` the router reads **`scope.ref` and nothing else**, matches it against the interest's `base_branches`, and wakes the worker. It never reads `before`/`after` and never compares against a stored head.
- It sets **no `wakeStateKey`** (only `check_run`/`check_suite` do, `:1510`), so `suppress_identical_wakes` (`:2857`) **never applies to push**. Every arriving push edge is a distinct wake, deliberately.

So what the consumer consumes is the **arrival**, not the column's value, and a last-state projection cannot carry an arrival. Under shadow that is a parity gap (101 of 138 unmatched smee events, CTL-48). **Under enforce with smee suppressed it is a wake that never happens, with no second copy to recover it from.** Hence the exclusion, until **CTC-704** keys `pushes` per delivery — at which point `PUSH_IS_LOSSY` goes false in the source and the set empties itself with no edit here.

## ⚠️ Readiness crosses a PROCESS boundary, and the naive port fails silently in the direction that looks fine

The Linear gate's `isReady` is a closure over the producer's timer (`daemon.mjs:1470`) — legitimate there, because `cloud-feed-timer` and `monitor.mjs` are both in the daemon. **The `github.*` consumer is `broker/tailer.mjs` → `broker/router.mjs`, a separate process.**

Copying the Linear wiring gives the broker `isReady: null`, and `decideDispatch`'s "absent probe ⇒ NOT ready" then makes smee authoritative **forever**. No double-dispatch, no error, no alarm — the host logs `mode: enforce` and behaves exactly like shadow. **The flip would be a lie**, and the way anyone would find out is by retiring the tunnel and losing dispatch. A safe fail direction is not a correct one.

`github-feed-ready.mjs` carries the verdict through a file, with a **staleness bound derived from the tick interval** — because a file without one is a latch: a daemon that dies mid-enforce leaves `ready: true` behind it, and the broker would keep suppressing smee on the authority of a process that no longer exists.

## Files

| file | |
| --- | --- |
| `github-feed-gate.mjs` | the decision, pure — per-name suppression, refused on **both** sides |
| `github-feed-ready.mjs` | the cross-process readiness signal + staleness bound |
| `github-feed-gate-install.mjs` | the broker-side gate; 1 s readiness cache (the broker routes **every** event in the fleet, 687/5 min into an 883 MB log — a read per event is not acceptable) |
| `github-feed-timer.mjs` | enforce emits **real names only for what the gate can suppress**, stamped with emission-time authority (CTL-1901); ready file written every tick, including un-ready ones |
| `broker/{tailer,index}.mjs` | gate decides before the router; armed before the tail starts |

The producer and the gate both read `GITHUB_SUPPRESSIBLE_NAMES`, so they cannot disagree — a disagreement is either a double dispatch or a dropped edge.

## Tests — 11 mutations run, and **3 of them passed**

New: 33 gate + 13 readiness + 12 installer + 4 wiring. Full `github-feed`/`cloud-feed` subset: **365 pass, 0 fail**.

⛔ **Three of my own tests were vacuous and are called out at their site**, because the mutation was run rather than reasoned about:

1. **The order control** asserted "readiness first would let a ready producer suppress `pr.merged`" — false. The suppress bit is invariant under that reordering; the block just runs later and catches it. Rewritten to assert what the order actually buys: **diagnosis**. An operator mid-cutover must distinguish "smee kept authority because the producer stalled" (transient) from "because this name has no replacement" (permanent until CTC-691).
2. **`PUSH_IS_LOSSY` tracking** compared the constant against `PUSH_IS_LOSSY ? [...] : []` — a tautology, since both sides are the same literal today. It would still pass after CTC-704, when the set is supposed to be **empty**. The derivation is now an injectable pure function driven in both states.
3. **"the gate is not ready before any read"** — the seeded cache value is unreachable, since the first probe always re-reads. Replaced with the property that makes it unreachable: construction does not probe, and the first probe does.

The same trap bit the `resolveEffectiveMode` counts ("9 of 12" hand-written passes against the derived value) — now driven with a different name set.

⭐ And the **smoke test caught a real defect before any test did**: taking the producer's emit-list as the gate's universe made `decideDispatch` answer `not-dispatch-class` for `github.pr.merged` — "the gate has no opinion" about the most dangerous name here. Safe today by accident; it would have recorded that the gate never looked rather than that it looked and refused. The universe is now what the **router** consumes (`GITHUB_CONSUMED_NAMES`).

## ⚠️ Gate run — not clean, and here is the accounting

Full `execution-core` suite on this branch: **9869 pass / 65 fail / 3 errors**. Baseline on `origin/main` (`6a413b301`), same machine, same run shape: **9803 pass / 61 fail / 3 errors**.

Diffed by test name, the delta is **4 tests, all in `__tests__/cloud-sync-idle-event.test.mjs`**:

- they **pass 5/5 in isolation** on this branch;
- they failed with `exitCode: null` — the spawned subprocess was **killed**, not an assertion mismatch;
- the file spawns `cloud-sync.mjs` with `CATALYST_REPLICA_START_TIMEOUT_MS=1` under a 9,935-test parallel run, with backend's schema cascade also running on this box;
- **`cloud-sync.mjs` imports nothing in this diff** — verified: not `github-feed-*`, not `daemon.mjs`.

That is the EAGAIN/resource-contention class, not a regression. Stating it rather than reporting a clean gate. **CI is authoritative here.**

## Rollout

Merged ≠ flipped. The flip is a separate, deliberate step: mini-2 first as canary under the tunnel-retirement runbook (CTL-1929), then mini, and **only after** schema 0.1.17 lands and the parity ledger returns CLEAN. Webhook `616654741` stays.

Refs CTL-1929, CTC-691, CTC-667, CTC-704, CTL-48, CTL-1901

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -1002,6 +1002,10 @@ jobs:
             github-feed-sweep.test.mjs \
             github-feed-parity.test.mjs \
             github-feed-timer.test.mjs \
+            github-feed-gate.test.mjs \
+            github-feed-ready.test.mjs \
+            github-feed-gate-install.test.mjs \
+            github-feed-gate-wiring.test.mjs \
             cloud-feed-gate.test.mjs \
             cloud-feed-capture.test.mjs \
             cloud-feed-timer.test.mjs \
--- a/plugins/dev/scripts/broker/index.mjs
+++ b/plugins/dev/scripts/broker/index.mjs
@@ -83,7 +83,7 @@ import {
   clearDebounceTimers,
   seedLastSeenByService,
 } from "./router.mjs";
-import { seedTailer, startTailing, stopTailing, loadExistingRegistrations } from "./tailer.mjs";
+import { seedTailer, startTailing, stopTailing, loadExistingRegistrations, initGithubFeedGate } from "./tailer.mjs";
 import { startCacheReconcileTimer } from "./cache-reconcile.mjs";
 import { gcStaleInterests } from "./gc-startup.mjs";
 import { getExecutionCoreDir, getRunsRoot } from "../execution-core/config.mjs";
@@ -531,6 +531,13 @@ function main() {
     },
   });
 
+  // CTL-1929: install the `github.*` dispatch gate BEFORE the tail starts.
+  //
+  // ⛔ ORDER IS LOAD-BEARING, same as the Linear gate's placement before monitorFn
+  // (daemon.mjs:1448). Arming it after `startTailing()` leaves a window in which
+  // enforce-mode smee events route normally — exactly the double-dispatch the gate
+  // exists to prevent, for however many events land in that window.
+  initGithubFeedGate();
   startTailing();
   const watchdogId = startWatchdog();
   const driftCheckId = startDriftCheckWatcher();
--- a/plugins/dev/scripts/broker/tailer.mjs
+++ b/plugins/dev/scripts/broker/tailer.mjs
@@ -25,6 +25,11 @@ import {
   isOrchestratorStatusFresh,
 } from "./router.mjs";
 import { getInterests } from "./state.mjs";
+// CTL-1929: the `github.*` dispatch gate. Imported from execution-core for the same
+// reason event-tail.mjs is — the two processes share this feature's modules, and the
+// gate must be the SAME decision function the producer's tests drive.
+import { decideDispatch as decideGithubDispatch } from "../execution-core/github-feed-gate.mjs";
+import { createGithubFeedGate } from "../execution-core/github-feed-gate-install.mjs";
 
 // Identity-stable alias — loadExistingRegistrations reports interests.size.
 const interests = getInterests();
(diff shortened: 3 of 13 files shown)


ryanrozich:
**No review arrived — Codex is out of quota** (COORD-136/137). Stating it per the standing rule.

This PR is **not on the rehearsal path**: it changes no host, and `CATALYST_GITHUB_FEED` stays `off` everywhere (`shadow` on mini-2 only). Merging on green under COORD-137's "non-rehearsal PRs merge on green with 'no review arrived' stated".

In place of the review round: **11 mutations run** against the new tests. **3 passed** — i.e. three of my own tests asserted nothing — and each is rewritten with the reason recorded at its site (the vacuous order control, the `PUSH_IS_LOSSY` tautology, and the unreachable cache seed). Quota ask filed as **CTL-1954**.

PR #3540, opened 2026-08-18
feat(dev): CTL-1929 — the installer stops opening the GitHub smee tunnel once the feed is authoritative

ryanrozich:
⚠️ **Stacked on #3535** (`ctl-1929-github-enforce-flip`) — it introduces the coverage lists doctor reads here. Base will retarget to `main` when #3535 merges.

## What this is

Step 3 of the tunnel-retirement runbook (posted on CTL-1929): **new** hosts stop creating the GitHub smee tunnel when `CATALYST_GITHUB_FEED=enforce`, and `catalyst doctor` stops reporting a declared-cloud node as having no GitHub replacement when it has one.

⛔ **Running tunnels are untouched.** That is what makes this landable before any flip, and it is why it is step 3 and not step 4 — step 4 (stopping the running client) is a separate, reversible action on each host.

## ⛔ `enforce` only — never `shadow`

In shadow the producer emits nothing authoritative and the dispatch gate suppresses nothing. Closing the tunnel there takes GitHub ingestion to **zero** on a host that only meant to observe — and the parity ledger would then be comparing the feed against an empty smee stream and **reporting agreement**. A silent, self-confirming failure.

So the call site reads `githubFeedIsAuthoritative()` rather than comparing a mode string inline: the next person to look at it should not be able to widen it to `!== "off"` without going to the function that names the rule. It ships with an explicit inversion guard, the same one CTL-1617's sibling gate documents (design §10 risk 7).

⛔ **It gates the GitHub tunnel only.** The two legs have separate knobs and separate rollouts by design; Linear's retired on its own terms (CTL-1928). A gate that closed both would couple two cutovers that were deliberately decoupled — covered by its own test.

## Doctor: a PASS only when nothing is left uncovered

`doctor.mjs:1325`'s declared-cloud WARN — *"GitHub ingestion has NO cloud replacement yet"* — is no longer unconditional:

| feed | grade |
| --- | --- |
| `off` | WARN — no source for `github.*` at all |
| `shadow` | WARN — **shadow replaces nothing** |
| `enforce`, gaps open | WARN, **naming the uncovered names** and the coverage ratio |
| `enforce`, all covered | PASS |
| resolver throws | WARN, fails closed to the `off` wording |

⛔ **Enforce is deliberately not an unconditional PASS.** A declared-cloud node has no tunnel at all, so `github.pr.merged` (CTC-691), `github.check_suite.completed` (CTC-667 item 4) and `github.push` (CTC-704) are genuinely unseen there. Reporting the route whole would say "fine" at exactly the moment the merge→deploy join, the CI wait and rebase detection are the parts that are missing. The counts and names are **derived**, so the WARN stops warning by itself when the gaps close.

## Two zero-import leaves — and a stale comment I repeated before measuring

**`lib/github-feed-names.mjs`** — the consumed/suppressible name lists, moved out of `github-feed-gate.mjs`. Doctor is now a consumer and runs under **bare Node**, which cannot load the gate (it reaches `bun:sqlite` via `github-feed-source.mjs`). The alternative was a hand-maintained copy in doctor that would keep reporting three uncovered names after the tickets close — the exact drift this feature keeps designing against.

**`lib/github-feed-mode.mjs`** — the canonical `CATALYST_GITHUB_FEED` resolver, with `config.mjs` **delegating** so four readers in three processes cannot disagree. A producer that read `enforce` while the tunnel gate read something else is a double dispatch on every covered name, and it would present as a parity failure rather than as the config drift it is.

⚠️ **I justified that second leaf with `deployment-mode.mjs`'s claim** that `config.mjs` is unloadable under bare Node — then measured it:

```
node -e "import('.../execution-core/config.mjs')"            → LOADS
node -e "import('.../execution-core/github-feed-gate.mjs')"   → REJECTS
node -e "import('.../execution-core/github-feed-source.mjs')" → REJECTS
```

**That comment is stale** — the sqlite reach became lazy at some point. The names leaf is load-bearing; the mode leaf is a consistency choice, and both headers now say so. A comment asserting another module's load behaviour is a measurement with no test behind it.

⚠️ Delegating also **corrects one behaviour**: a set-but-invalid env value now falls back to `off` **and overrides Layer-2**, per CAT-57's documented contract (`config.test.mjs`: *"a SET-but-invalid env value ... overrides Layer-2 enforce"*), which `readGithubFeedConfig` did not follow. Empty/whitespace still defers to Layer-2 — unset and `""` are the same operator intent, and treating `""` as a typo would let `export CATALYST_GITHUB_FEED=` silently disable a Layer-2 rollout.

## Tests

New: **17** leaf + **5** doctor + **4** tunnel-gate. Affected suites together: **922 pass, 0 fail**.

**11 mutations run.** Two of my own tests were vacuous and are rewritten with the reason recorded at their site:

1. **The fail-closed assertion on a throwing resolver passed with the mutation applied** — while the three gaps are open, *both* branches grade WARN and both print the source, so `status === WARN` could not distinguish fail-open from fail-closed. It would have started being wrong precisely when the gaps closed and fail-open became a PASS. Now asserts *which branch answered*.
2. The doctor coverage detail, same class as the ones called out in #3535.

⭐ **And `config-pino-fallback.test.mjs` caught a real integration gap on the first full-gate run**: its fixture stages every relative import `config.mjs` has, and my new leaf was not staged, so the module failed to load under the no-deps fixture. Its own comment states that contract — the test worked exactly as designed, and the fix is one `cpSync`.

## ⚠️ Gate accounting

Full execution-core suite: **9893 pass / 63 fail** before the pino fix. Diffed by name against an untouched `origin/main` worktree (58 failures) and against #3535's branch, the **only** delta introduced by this commit was those 2 pino-fallback tests — now fixed and green. The rest are the same pre-existing local-environment failures documented on #3535 (hostname, replica, scheduler, and the `cloud-sync-idle-event` EAGAIN class).

Separately, 4 pre-existing failures in `deployment-mode-tunnel-gate.test.ts` are ~6 s `fetch` timeouts local to this machine — **identical 4 pass / 4 fail on untouched main**. My tunnel tests assert on the tunnel constructor rather than through HTTP for that reason.

**CI is authoritative.**

Refs CTL-1929, CTC-691, CTC-667, CTC-704, CTL-1928, CAT-57, CTL-1617

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -1041,6 +1041,7 @@ jobs:
             github-feed-ready.test.mjs \
             github-feed-gate-install.test.mjs \
             github-feed-gate-wiring.test.mjs \
+            github-feed-mode-leaf.test.mjs \
             cloud-feed-gate.test.mjs \
             cloud-feed-capture.test.mjs \
             cloud-feed-timer.test.mjs \
--- a/plugins/dev/scripts/execution-core/config-pino-fallback.test.mjs
+++ b/plugins/dev/scripts/execution-core/config-pino-fallback.test.mjs
@@ -58,6 +58,11 @@ function stageScratch() {
   // module graph resolves identically to production (same rationale as deployment-mode.mjs
   // above).
   cpSync(resolve(__dirname, "../lib/secret-contract.mjs"), join(libDir, "secret-contract.mjs"));
+  // CTL-1929: readGithubFeedConfig now delegates to the zero-import github-feed-mode
+  // leaf, so it is part of config.mjs's relative-import graph too. ⭐ This test caught
+  // the omission on the first full run rather than letting a host discover it — which
+  // is exactly the contract the comment above states, working.
+  cpSync(resolve(__dirname, "../lib/github-feed-mode.mjs"), join(libDir, "github-feed-mode.mjs"));
   // type:module + no deps -> pino unresolvable from this directory tree.
   writeFileSync(
     join(scratch, "package.json"),
--- a/plugins/dev/scripts/execution-core/config.mjs
+++ b/plugins/dev/scripts/execution-core/config.mjs
@@ -85,6 +85,10 @@ export {
   resolveDeploymentMode,
   getDeploymentMode,
 } from "../lib/deployment-mode.mjs";
+// CTL-1929: the canonical github-feed resolver. Zero-import leaf for the same reason
+// deployment-mode.mjs above is one — doctor.mjs runs under bare Node and cannot load
+// this file (its chain reaches bun:sqlite via linear-query.mjs).
+import { resolveGithubFeedMode } from "../lib/github-feed-mode.mjs";
 
 // --- Paths ---
 // Re-resolved per call so tests can redirect by setting CATALYST_DIR;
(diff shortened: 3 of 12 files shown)


ryanrozich:
**Rebased onto `main` and base retargeted** — #3535 merged as `98ec1ed` (ancestry-verified), so this is no longer stacked.

**No review arrived — Codex is out of quota** (COORD-136/137, ask filed as **CTL-1954**). Not on the rehearsal path: it changes no running host, and the tunnel gate is inert unless `CATALYST_GITHUB_FEED=enforce`, which is set nowhere. Merging on green under COORD-137.

In place of the review round: **11 mutations run**, **2 passed** and are rewritten with the reason at their site — most usefully the fail-closed assertion on a throwing resolver, which could not distinguish fail-open from fail-closed while the three coverage gaps are open (both branches grade WARN) and would have started being wrong exactly when the gaps closed.

PR #3544, opened 2026-08-18
feat(dev): CTL-1929 — wire pr.merged and per-push into the producer on schema 0.1.17

ryanrozich:
Schema **0.1.17 is pinned and live on mini-2** — verified by content rather than by waiting for the rollout turn: `pull_requests.merge_commit_sha` present, `push_events` keyed on `delivery_id`, `check_suites` populating.

Two of the three blocking names are now wired. **The third is not, and the reason changed** — see below.

## ⭐ `github.pr.merged` emits (CTC-691)

The stream was deliberately kept paging while the column was missing, so CTC-691 would land as a **deletion** here rather than as new read-layer code. It did: `UNCOVERED_STREAM_REASONS` is now empty.

⛔ **The whole-stream refusal became a per-ROW one**, which is the only correct granularity. CTC-691 added the column **without backfilling** the ~4,230 pre-existing PR rows (COORD-124), so every merge predating the pin has it NULL. Those decline with `no-merge-commit-sha:not-backfilled`, visibly, in `byReason`.

That guard is not fastidiousness. `mergeCommitSha` is the **join key**: `router.mjs:1513` writes it via `setFilterStateMerged` and `github.deployment.created` later matches `WHERE merge_commit_sha = ?`. A twin without it **routes and wakes `monitor-merge` normally** while `monitor-deploy` stops firing — it looks exactly like success. Same posture `deploymentCreated` already takes for its own sha.

## ⭐ `github.push` comes from `push_events` (CTC-704)

One row per delivery, PK = GitHub's `x-github-delivery`. That makes it an ordinary append-only stream: no `mutableRow`, no folded coordinate, and a redelivery dedups for free because GitHub reuses the id.

⛔ **Both push streams carry the same event name, so `availableStreams` resolves the replica down to EXACTLY ONE.**

- Running **both** on a pinned host emits every push twice — and their identities differ (`delivery_id` vs the folded `repo_id@ref`), so the seen-set *cannot* collapse them: the duplicate reaches the router as a second, genuine-looking base-branch move and wakes every waiter again.
- Running **neither** on an un-pinned host silently kills rebase detection.

⚠️ This is a **live condition tonight, not a hypothetical**: the canary leaves one host on the old pin by design (COORD-128 §2). So `tableExists` asks the database rather than inferring from a version, and `PUSH_IS_LOSSY` became `pushIsLossy(db)` — a module constant is wrong on exactly one host of a mixed fleet, and the dispatch gate reads it to decide whether `github.push` may suppress smee.

## ⛔ `check_suite.completed` still does NOT emit — and the reason is not the old one

The table landed in 0.1.17. **The association the consumer keys on did not.**

`router.mjs:1497` reaches an interest **only** through `detail.prNumbers`, which the webhook fills from `check_suite.pull_requests[].number`. The mirror drops that field, and `check_runs` cannot supply it (also PR-less). The only available derivation is a **last-state** join through `pull_requests.head_sha`:

```
check_suites total = 202 · head_sha joins a PR = 93 (46%) · misses = 109
misses by branch: ryan/ctl-1944-direnv-fleet 25 · ryan/ctl-1949-ask-traps 19
                  ryan/ctl-1952-session-verdict 10 · ctc-pin-sdk-0.8.13 10
                  main + release-please 43 (legitimately no PR)
```

⭐ The sharpest row: **`ctc-pin-sdk-0.8.13` appears 10 times in the hits and 10 in the misses** — one branch, where only suites matching the PR's *current* head resolve. That is the superseded-head pattern isolated, and it is why 46% is structural rather than a sampling artifact.

Emitting on a 46% join would hang the CI wait silently, with no smee copy under `enforce`. **→ CTC-712, filed P1.** So `github.check_suite.completed` stays in `GITHUB_UNCOVERED_NAMES`, with the entry rewritten to name the real cause.

## The ledger

`github.pr.merged` gains a compare spec — including `mergeCommitSha`, with a note that it must never move to `OBSERVED_ONLY` — and **leaves `KNOWN_ABSENT`**.

⛔ That removal is the point of CTC-691, not a side effect. While `pr.merged` sat in `KNOWN_ABSENT` the ledger **excused its absence** — so the one name whose loss silently kills the merge→deploy chain was precisely the one this instrument was configured not to notice. A ledger that cannot fail on a name is not measuring it. There is now a test asserting its absence breaks `clean`, plus the positive control that a matched pair still compares clean.

## Tests

**238 pass / 0 fail** across the `github-feed` suite (+21 new). **8 mutations run, all 8 fired:**

| | |
| --- | --- |
| supersession removed (both push streams run) | 2 fail |
| table existence assumed | 1 fail |
| `pushIsLossy` always true | 1 fail |
| un-backfilled rows emit a twin with no join key | 5 fail |
| payload says `merged: false` | 1 fail |
| join key dropped from the payload | 1 fail |
| invents `action: "merged"` (GitHub never sends it) | 1 fail |
| sweep defaults to `STREAMS` instead of asking the replica | 1 fail |

New coverage includes the force-push-back-to-a-previous-sha case (which collapses under COORD-127's originally proposed `(repo_id, ref, after)` key, and is why backend's delivery-id choice was right), and a control that the two push streams produce **byte-identical envelopes** — they differ in which rows they yield, never in what an event looks like.

## ⚠️ Gate accounting

**Full gate deferred: load average was 172 at commit time**, against a rule of 40. Running it would have produced a result I could not trust and would have made the box worse for everyone else mid-rollout. Targeted suite is 238/0. **CI is authoritative.**

## What this does NOT do

⛔ Changes no host. `CATALYST_GITHUB_FEED` is `off` everywhere except mini-2 (`shadow`). The tunnel does not retire until the ledger returns CLEAN over Tuesday-morning traffic **and** CTC-712 closes — precondition P2 on the CTL-1929 runbook stays open.

⚠️ The gate's suppressible set still excludes `github.push` (it derives from the deprecated static `PUSH_IS_LOSSY`, which reports the pre-CTC-704 world). That is the **safe** direction — smee keeps authority for push, exactly one copy dispatches — and resolving it per-host is a follow-up, not something to guess at here.

Refs CTL-1929, CTC-691, CTC-704, CTC-712, CTC-667, COORD-124, COORD-128

### DIFF
--- a/plugins/dev/scripts/execution-core/github-feed-event.mjs
+++ b/plugins/dev/scripts/execution-core/github-feed-event.mjs
@@ -76,6 +76,10 @@
 
 import { buildCanonicalEvent } from "./lib/canonical-event.mjs";
 import { STREAM_BY_KEY } from "./github-feed-source.mjs";
+import {
+  GITHUB_CONSUMED_NAMES,
+  GITHUB_UNCOVERED_NAMES as LEAF_UNCOVERED_NAMES,
+} from "../lib/github-feed-names.mjs";
 
 /** The service identity every `github.*` event must carry. See the header. */
 export const GITHUB_SERVICE_NAME = "catalyst.github";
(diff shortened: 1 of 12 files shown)


ryanrozich:
⛔ **CI caught real drift that passed on both branches locally — pushed a fix.**

#3540 moved the consumed/uncovered name lists to the zero-import leaf `lib/github-feed-names.mjs` (doctor runs under bare Node and cannot load `github-feed-gate.mjs`) and pointed the **gate** at the leaf — but left `github-feed-event.mjs`'s own copies in place. This branch then removed `github.pr.merged` from **the copy the gate no longer reads**.

Each branch passed locally because each predated the other's change. Rebased onto main the two lists disagreed, and in the bad direction: the gate kept **excluding** `github.pr.merged` while the producer had started emitting it under its real name, so the feed's authority on that name was silently void. (Not a double dispatch — exactly one copy still dispatches, smee's — but the cutover would have looked done and not been.)

⭐ That is exactly the drift the leaf exists to prevent, arriving through a half-finished migration rather than through a hand-maintained list. **Fixed at the shape rather than at the site:** `github-feed-event.mjs` now DERIVES from the leaf (`GITHUB_DISPATCH_CLASS_NAMES` = consumed − uncovered) instead of restating it, so there is one array and identity holds across producer, gate and doctor.

New test asserts that identity with `toBe`, not `toEqual` — a second array with equal contents is precisely what is being guarded against. Mutation (re-adding `pr.merged` to the leaf) fails 3 tests. Suite: **699 pass / 0 fail** across `github-feed` + `doctor`.

⚠️ Worth noting for the record: **this is the failure Codex would most likely have caught** on #3540, and it is out of quota tonight (CTL-1954). CI caught it instead, which is the backstop working — but one round later than a reviewer would have.

PR #3551, opened 2026-08-18
fix(dev): CTL-1929 — the parity ledger compares only repos smee can deliver, so CLEAN is reachable

ryanrozich:
⚠️ **Stacked on #3544.** Base retargets to `main` when that merges.

## ⛔ The cutover gate was unmeetable for a reason that was not a fault

Measured over a **55-minute live window** on mini-2, through the shipped ledger:

```
joined 81 / agree 81 · smee-unjoined 0 · feed-unjoined 6
clean = false · exit 3   ← blocked on "feed-events-without-a-twin: 6"
```

**81 of 81 joined events agreed on every compared field, and not one smee event lacked a twin.** The verdict was still INCONCLUSIVE. All six unjoined feed events came from repos the tunnel does not deliver:

| | repos |
| --- | --- |
| **smee delivers** | `catalyst` · `catalyst-cloud` · `ryanrozich/personal-os` |
| **feed produces** | `catalyst` · `catalyst-cloud` · **`catalyst-cloud-sdk`** · **`thoughts`** |

The webhook subscription covers a repo set an operator configured; the mirror ingests a different, larger one.

⭐ **So the feed is BROADER than the tunnel, and retiring it EXTENDS coverage rather than shrinking it.** Backend's own SDK PR #26 — the 0.8.13 leg of tonight's cascade — was visible to the feed and never to the tunnel.

⛔ But `feed-events-without-a-twin` was conflating two things that need **opposite** verdicts:

- the feed emitted a **phantom** → a real defect, must block CLEAN;
- the feed covers a repo smee **cannot** → correct behaviour, and it blocked CLEAN **forever**.

That is the mirror image of the false-clean corrected in CTL-48 — the same instrument, wrong in the other direction, and just as capable of driving a wrong decision. A gate that can never be met is not a safety property; it is a gate someone eventually overrides.

## The fix

The comparison is scoped to the repos **smee actually delivered in the window**. Feed events outside that set are reported as `feedOnlyRepos` and `comparableRepos` — **never dropped**, because an operator reading a clean verdict has to see that the feed covered more. That is a reason to retire, not a caveat on retiring.

⚠️ **The scope is derived from the SMEE side only.** Taking the intersection of both sides would also excuse a repo the feed *should* cover and silently does not — a real defect that must keep failing. Driven by its own test.

⚠️ **An event with no repo attribute stays COMPARABLE.** Letting a malformed envelope excuse itself out of the comparison is the shape of every false-clean in this file's history.

## Verified against the window that motivated it

```
before: feed-unjoined 6 · clean=false
after : feed-unjoined 0 · feedOnlyRepos {catalyst-cloud-sdk: 5, thoughts: 1}
                         comparableRepos [catalyst, catalyst-cloud]
```

⭐ **The path to CLEAN is now fully visible**, and every remaining blocker is known and tracked:

1. `github.pr.merged` — 14 unexplained absences, because mini-2 has not yet pulled **#3544**. Once it does, the producer emits them and they join.
2. `github.check_suite.completed` — **CTC-712** (P1, one column: the `pull_requests` association the consumer keys on).

Nothing else stands between the current state and `clean = true`.

## Tests

**30 pass / 0 fail** in the parity suite (+6 new). **4 mutations run, all fired:**

| | |
| --- | --- |
| no-repo events excused as feed-only | 1 fail |
| scoping computed but not applied | 1 fail |
| feed-only repos silently dropped from the report | 2 fail |
| scope = union of both sides (hides a genuine feed gap) | 2 fail |

The load-bearing control is *"a feed event from a **shared** repo with no twin STILL blocks clean"* — the scoping must excuse only what smee could never have produced, never a genuine phantom in a repo it covers.

Refs CTL-1929, CTL-48, CTC-712

### DIFF
--- a/plugins/dev/scripts/execution-core/github-feed-event.mjs
+++ b/plugins/dev/scripts/execution-core/github-feed-event.mjs
@@ -524,11 +524,16 @@ export function buildGithubEvent(streamKey, row, seams) {
         entity: "pr",
         action: "merged",
         label: `PR #${row.number}`,
+        // ⛔ NO `vcs.revision`, THOUGH THE MERGE SHA IS RIGHT THERE. I added it — it is
+        // the obvious attribute for a merge commit — and the parity ledger caught it
+        // on the first live window: the WEBHOOK does not set it on this name, so the
+        // feed's copy diverged on all 3 events. Reproduce the webhook, do not improve
+        // it; the same rule that keeps `threadId: 0` on pr_review_thread.resolved.
+        // The consumer reads `detail.mergeCommitSha` from the payload, which is
+        // populated below.
         attrs: {
           "vcs.repository.name": repo,
           "vcs.pr.number": row.number,
-          // The merge commit, on the attribute the deploy chain scopes by.
-          "vcs.revision": row.merge_commit_sha,
         },
         message: `${name} for ${repo} PR #${row.number}`,
         payload: {
@@ -538,7 +543,14 @@ export function buildGithubEvent(streamKey, row, seams) {
           // read the pair, so inventing an action nobody emits would leave the
           // lifecycle branch unmatched while every count still read "emitted".
           merged: true,
-          mergedAt: typeof row.merged_at === "number" ? new Date(row.merged_at).toISOString() : null,
+          // ⛔ SECOND-PRECISION, NOT MILLISECOND. `toISOString()` always emits `.000Z`;
+          // GitHub's own string omits milliseconds, so the naive form diverged on
+          // every event (`08:07:22.000Z` vs `08:07:22Z`) — caught by the ledger on the
+          // first live window, exactly where #3544's own PR body predicted it would go
+          // first. Trimming is safe because the column is whole seconds from GitHub.
+          mergedAt: typeof row.merged_at === "number"
+            ? new Date(row.merged_at).toISOString().replace(/\.\d{3}Z$/, "Z")
+            : null,
           // The join key. `classifyGithubRow` has already refused the row if it is
           // absent, so this is never null here — the guard is there, not here.
           mergeCommitSha: row.merge_commit_sha,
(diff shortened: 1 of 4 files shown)


ryanrozich:
Rebased onto `main` and **base retargeted** — #3544 merged as `538214a` (ancestry-verified), so this is no longer stacked and now carries a single commit.

⚠️ The rebase conflicted because #3544 was **squash**-merged: this branch still held the pre-squash commits, so the wire change looked like an unapplied delta against a main that already contained it. Reset to `main` and cherry-picked the one commit that is actually mine here. Suite after the reset: **263 pass / 0 fail**.

**No review arrived — Codex is out of quota** (CTL-1954). 4 mutations run in its place, all fired.

ryanrozich:
⭐ **Second commit added — the ledger caught two real divergences in my own `pr.merged` envelope**, on the first live window after #3544 landed, on all 3 events.

| field | feed | smee |
| --- | --- | --- |
| `attributes.vcs.revision` | `52c90b8c…` | **absent** |
| `body.payload.mergedAt` | `2026-08-18T08:07:22**.000**Z` | `2026-08-18T08:07:22Z` |

⛔ **1.** I set `vcs.revision` to the merge sha because that is the obvious attribute for a merge commit. **The webhook does not set it on this name.** Removed from both the envelope and the compare spec — *reproduce the webhook, do not improve it*, the same rule that keeps `threadId: 0` on `pr_review_thread.resolved` even though the replica has a real id. The consumer reads `detail.mergeCommitSha` from the payload, which is untouched.

⛔ **2.** `toISOString()` always emits `.000Z`; GitHub's string omits milliseconds. Trimmed to second precision (safe — the column is whole seconds from GitHub).

⭐ Worth recording: **#3544's own PR body predicted `mergedAt` would be the field most likely to diverge first, and argued against pre-excluding it** so the ledger would say so loudly. It diverged within minutes of landing, and finding it was worth more than the exclusion would have been. Both are the same class as the four traps in the CTL-1929 handoff — *a field that is nicer than the webhook's is still a divergence*, and the instrument the cutover is judged by is the thing it breaks.

Suite: **263 pass / 0 fail**.

PR #3552, opened 2026-08-18
fix(dev): CTL-1956 — the direnv provisioning a host actually needs, on hosts that have no brew

ryanrozich:
## The measurement this starts from

CTL-1944 shipped the readiness check and `ensure_direnv`. Measured on **mini-2** today, *after*
that fix had reached the host (`plugin-source` at `faaddb1a`):

```
direnv fleet readiness — host mini-2, dir ~/code-repos/github/coalesce-labs/catalyst-cloud
  ❌  direnv NOT installed — this host cannot materialize fleet tokens (install: brew install direnv)
FAILED: 1   EXIT=1
```

⭐ The **check** is right — that is a real red, not a synthetic mutation. The **installer** is what
was inert:

```sh
if ! command -v brew >/dev/null 2>&1; then
  echo "  warning: direnv absent and brew unavailable — install manually…" >&2
  return 0     # ← the only remediation path was brew
fi
```

⛔ **Neither fleet mini has Homebrew.** On the two hosts the ticket was written about, the codified
installer installed nothing. mini's own direnv lives at `~/.local/bin` because somebody typed it —
precisely the uncodified machine edit CTL-1908 is about. This is the CTL-1919/CTL-1935 class: the
fix is real, the *delivery* is not.

## And the binary was 1 of 4 missing pieces — the check saw one

| piece | laptop / mini | mini-2 | check saw it? |
| -- | -- | -- | -- |
| `direnv` binary | ✅ | ❌ | ✅ |
| `lib/profiles.sh` — **defines `use_profile`** | ✅ | ❌ | ⛔ **no** |
| `lib/otel.sh` — defines `use_otel_context` | ✅ | ❌ | ⛔ **no** |
| `coalesce-labs/.envrc` — the `source_up` target | ✅ | ❌ | ⛔ **no** |
| profiles `personal.env` / `catalyst.env` | ✅ | ❌ | ✅ |

⚠️ `profiles.sh` was **in no repository at all** — byte-identical on two machines
(`md5 240be047…`) only because it had been copied there. `use_profile` is not a direnv builtin, so
without it every `.envrc` dies on an undefined function and **every token reads EMPTY**. The check
then went red naming three empty tokens instead of the one missing helper that caused them.

## What changed

- **Vendored `plugins/dev/direnv/lib/profiles.sh`** (secret-free — the credentials live in the
  `profiles/` files it sources), beside the already-vendored `otel.sh` that nothing installed.
- **`ensure_direnv` falls back to direnv's bare static release binary** → `~/.local/bin` when brew
  is absent, **pinned** to the version the fleet runs: an unpinned upgrade is exactly how a host
  would acquire a third `Found RC allowed` encoding the check fails closed on. The download must
  **execute** before it takes the name — a truncated file is still a file and would read as
  installed to every `command -v` downstream.
- **`ensure_direnv_runtime`** provisions the rest: both lib helpers, the `source_up` target, and
  placeholder `personal`/`catalyst` profiles. ⛔ Every write is **create-if-absent** — mini's
  profiles are richer than any placeholder and clobbering them would break a working host to fix a
  broken one.
- **Both functions moved to `lib/direnv-provision.sh`**, sourced by *both* `install-cli.sh` and
  `check-direnv-fleet.sh --install`. Running the whole join stage is not a reasonable ask
  mid-rollout, so the narrow path has to be real — and there must not be two implementations to
  drift apart.
- The check **names a missing lib helper individually**, and `--install` now also runs
  `direnv allow` — then **re-reads** the state, so an allow that silently fails still renders RED.

## Verified on the actual host

Run on mini-2 through the codified path (not by hand):

```
  Installing direnv v2.37.1 → ~/.local/bin (no brew on this host — CTL-1956)…
  Installed direnv 2.37.1 (/Users/ryan/.local/bin/direnv)
  Installed direnv lib/profiles.sh → …   Installed direnv lib/otel.sh → …
  Created …/coalesce-labs/.envrc (source_up target)
  Created placeholder profile personal.env / catalyst.env
  ✅ direnv installed   ✅ lib profiles.sh   ✅ lib otel.sh   ✅ 3 profiles
  ℹ  .envrc BLOCKED — running: direnv allow …
  ✅ .envrc is ALLOWED (Found RC allowed 0, after --install)
  ✅ LINEAR_API_TOKEN (48)   ✅ GITHUB_TOKEN (40)   ✅ live Linear viewer probe OK
  ❌ CLOUDFLARE_API_TOKEN is EMPTY
```

⭐ A remote owner on mini-2 can now read Linear. The one remaining red is a **credential**, not
code — it came from the hand-provisioned `personal.env` mini-2 never had, and is fixed by
coalesce-labs/catalyst-cluster#7 (adds it to the cluster bundle).

## Tests

`check-direnv-fleet` **38/38** (was 28) · new `direnv-provision` suite **9/9**. Both wired into
`execution-core-tests.yml` — a suite that is not in CI ships inert, which is this ticket's own
failure mode.

Every new assertion has a control that fires:
- removing one lib helper must leave the **other** reported present (a check that reds both rows
  tells you nothing about which to fix);
- restoring both must return the *same call* to green;
- the no-clobber md5 comparison is proven able to detect a changed file.

⭐ **Mutation control:** extracting these functions into `lib/` silently re-pointed the vendored
lookup one level too high, so provisioning would have "run" and installed nothing. Re-introducing
that path takes the new suite to **6 pass / 3 fail** — the guard catches the exact defect.

Pre-existing and **not** from this PR: `install-cli.test.sh` is 67/70 on this branch and **67/70 on
`origin/main`**, with an *identical* failure set (`comm`-compared by name, not by count) — three
`wrapper-mode` version cases.

⚠️ `make test` (the full monorepo gate) was attempted once and **timed out at 10 min** with the
shared laptop at load 41; the directly-affected suites above were run instead. CI is authoritative.

Refs CTL-1944, CTL-1908.


ryanrozich:
⚠️ **Merged with NO review arrived** — `chatgpt-codex-connector[bot]` posted *"You have reached your Codex usage limits for code reviews"* rather than a review (the quota ask is CTL-1954). Per COORD-135/137 this is stated as **"no review arrived"**, not as a clean pass: zero threads is not an approval.

Not on the rehearsal path (host provisioning), so no peer read was requested. All 6 required checks green; verified on `main` by content afterwards — `lib/direnv-provision.sh`, the vendored `profiles.sh`, the new suite, and its CI wiring are all present at `457f0254d`.

⭐ Verified on the actual host after merge: **mini-2 `direnv fleet readiness: PASS`, exit 0, 12/12** (with coalesce-labs/catalyst-cluster#7 supplying the last credential).

PR #3555, opened 2026-08-18
feat(dev): CTL-1889 inc 3 — the cluster fence + heartbeat go through the cloud attachment proxy (CTC-692)

ryanrozich:
Increment 3 of **CTL-1889**, and the half of **CTC-692** that lives in this repo. Increment 1 moved `issue-state` + `label`, increment 2 moved comments; this moves the last two execution-core app-actor writers: `cluster-claim.mjs` (the fleet's cross-host soft-CAS mutex) and `cluster-heartbeat.mjs` (per-host liveness).

## ⛔ The entire risk of this change is that the proxy NEVER THROWS

`defaultPost` throws on a transport error, a non-2xx, or a GraphQL `errors[]` body. `proxy.send` / `proxy.read` return **tagged verdicts and never throw**. `cluster-claim.mjs` depends on that throw in two places where nothing else is watching:

**1. A failed READ swallowed into `[]` resets the fencing token.**
```js
const current = await readClaim(ticket);        // fails → []  → null
const nextGen = (current?.generation ?? 0) + 1; // 1, on a ticket held at generation 7
```
We then write our claim at generation 1, and the read-back confirms *our own write* → `won: true`. Two hosts own the ticket, `isFenceCurrent(ticket, 7)` starts answering false for the host actually doing the work, and **no error is raised anywhere.**

**2. A failed WRITE that does not throw invents a win.** The stale-preemption branch is:
```js
await writeClaim(...); return { won: true, generation: nextGen };   // ⛔ no read-back
```
`writeClaim`'s throw is the *only* thing between a refused write and a host believing it owns a ticket it never wrote to. A `{applied:false, reason:"budget:day-exhausted"}` handed back quietly turns this host's own write budget into a fleet-wide double dispatch.

So `cluster-attachment-transport.mjs` re-establishes **one invariant — every non-success throws, with a named reason** — and both mutation controls in its suite *reproduce the defect* when the invariant is removed, so the assertions cannot go vacuous.

## ⭐ Measured live, not assumed: no UUID resolution is needed

`cluster-claim.mjs`'s header said `attachmentCreate` "needs the UUID, not the identifier". Probed against the **live** route with a per-host org key:

| probe | result |
| -- | -- |
| `POST /agent/attachment {"issueId":"CTL-1889", …}` | `200 succeeded` — **the identifier is accepted** |
| same `url` again, new metadata | **same attachment id** `d9918a35…`, metadata replaced |
| `GET /agent/attachments?issueId=CTL-1889` | exactly **one** node at that url, metadata intact |

⭐ This matters beyond tidiness. The only credential-free way to resolve a UUID on-host is the freshness-gated replica resolver — which would have coupled **every cluster claim to the replica writer's heartbeat**, so a dead replica writer would have stopped the fleet claiming *any* ticket. Passing the identifier removes that coupling entirely. The stale header claim is corrected in place; `resolveIssueId` is kept for the direct path, where it has **not** been re-measured.

## The read-back is a THIRD peer, not a fourth write route

`/agent/attachments` is a GET and the cloud spends **no** write-budget unit on it. `read()` mirrors that — no budget, no convergence suppression — and both omissions are correctness requirements, not tidiness:

- ⛔ **if the read-back spent budget**, a host that exhausted its daily writes could no longer *verify* a claim → every claim returns `won:false` → **that host stops dispatching entirely**, a total work stoppage caused by a budget designed to bound writes;
- ⛔ **convergence** means "skip the call, the state is already right", which applied to a live fencing question would answer from a cache.

`buildProxyRequest` grew GET support, and `buildCurlConfig` now emits **no `data` line** for a bodyless request — curl treats any request carrying a body as a POST regardless of `request = "GET"`, so a stray `data = ""` would 404 on a route that exists. Asserted as an absence, with a negative control that the builder *does* emit `data` for a write.

## Live verification — real route, per-host org key

**Two concurrent claimers × 4 rounds → exactly one winner every time.** Generations strictly monotonic `1 → 2 → 3 → 4`, the fence always naming the winner, the loser always `won:false`:

```
host-a => {"won":false,"generation":4}
host-b => {"won":true,"generation":4}
fence  => {"owner_host":"probe-host-b","generation":4,…}
winners=1  RESULT: PASS
```

⚠️ This proves *exactly one winner*, **not fairness** — `host-b` won all four rounds because the harness's interleaving is deterministic.

**Negative controls — all three throw with a named reason**, never a silent `won:false` and never a fall-back to the app-actor:

| credential | result |
| -- | -- |
| none | `THREW reason=no-cloud-token` |
| the tenant-wide `ADMIN_TOKEN` | `THREW reason=unauthorized` (401) |
| a bogus key | `THREW reason=unauthorized` (401) |

And `claimDispatchSync` already maps a non-zero CLI exit to `{won:false}`, so a thrown claim reaches the scheduler as "did not win" — fail-closed end to end.

## ⚠️ Two suites that were never in CI, and one assertion that inverted

- **`cluster-heartbeat.test.mjs` and `cluster-heartbeat-capacity.test.mjs` were NOT in the workflow's hand-enumerated list.** The list is not globbed (the workflow says so), so a module this PR changes had **zero** CI coverage. Both are added, with the new suite.
- **`linear-write-proxy.test.mjs` used `"attachment"` as its example of an UNKNOWN route.** Making that route real inverted the assertion. It now uses an id that cannot become real — a plausible-sounding placeholder turns "we shipped the route you asked for" into a red test in an unrelated file.

## Verification

| | |
| -- | -- |
| targeted suites (9 files, incl. all cluster + proxy) | **364 pass · 0 fail** |
| new `cluster-attachment-transport.test.mjs` | 36 tests, incl. 2 mutation controls that reproduce the defect |
| live CAS, real route | 4/4 rounds, exactly one winner, monotonic generations |
| live negative controls | 3/3 throw with named reasons |

⚠️ **On the full local gate: it is NOT being reported as green, because it was invalidated mid-run.** A post-commit hook auto-rebased the branch onto `origin/main` while `run-tests.sh` was still executing, so the shell suites before and after that point ran against different trees. The targeted suites above were re-run to completion on the final rebased tree. **CI is the authority here.** Three suites were already red on clean `origin/main` before this branch existed, verified in a detached baseline worktree at `457f0254d` with byte-identical failing assertions: `catalyst-join.test.sh` (T3.4), `catalyst-legacy-plugin.test.sh` (version 1.0.0), `catalyst-stack.test.sh` (59/68).

## ⛔ Not armed on the fleet by this PR

Both minis are in `CATALYST_LINEAR_WRITE_PROXY=enforce`, so loading this code arms the new path immediately. It is deliberately **not** loaded: **#3548 (the 0.1.18 / sdk 0.8.14 fleet pin) is still open**, and the sequencing on CTL-1889 is pin-first.

The preflight for whoever loads it is measured and recorded on CTL-1889 — both hosts' **live** `CATALYST_CLOUD_TOKEN` (read from the running daemon's process env, not a config file) returns **HTTP 200** on `/agent/attachments`, so neither will lose the ability to claim. ⚠️ That also corrects a stale note in `linear-write-proxy.mjs`'s header saying mini still carries the `ADMIN_TOKEN`; it carries a 58-char per-host org key.

Refs CTL-1889, CTC-692.

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -989,6 +989,9 @@ jobs:
             checkout-sync.test.mjs \
             cluster-claim-sync.test.mjs \
             cluster-claim.test.mjs \
+            cluster-attachment-transport.test.mjs \
+            cluster-heartbeat.test.mjs \
+            cluster-heartbeat-capacity.test.mjs \
             comms-drain.test.mjs \
             codex-run-phase-agent.test.mjs \
             daemon-watchdog-alert.test.mjs \
--- a/plugins/dev/scripts/execution-core/cluster-attachment-transport.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-attachment-transport.mjs
@@ -1,0 +1,271 @@
+// cluster-attachment-transport.mjs — CTL-1889 increment 3 / CTC-692.
+//
+// The seam that lets `cluster-claim.mjs` (the fleet's soft-CAS mutex) and
+// `cluster-heartbeat.mjs` (per-host liveness) reach Linear's attachment verb EITHER
+// directly with this host's own app-actor (the pre-CTL-1889 path) OR through the Catalyst
+// Cloud write proxy under the ONE cloud-held grant. These are the last two app-actor
+// writers in execution-core; increments 1 and 2 moved the issue-field writes and comments.
+//
+// ══════════════════════════════════════════════════════════════════════════════════════
+// ⛔⛔ THE ONE INVARIANT THIS MODULE EXISTS TO ENFORCE: EVERY FAILURE THROWS.
+// ══════════════════════════════════════════════════════════════════════════════════════
+//
+// Not "returns null". Not "returns an empty list". Not "returns {ok:false}". THROWS.
+//
+// This is not defensive style — it is the difference between a mutex and a corruption,
+// and both callers depend on it in ways that are invisible at their call sites:
+//
+// ⛔ A READ THAT RETURNED `null`/`[]` ON FAILURE SILENTLY DOUBLE-CLAIMS. `claimTicket`
+//    opens with `const current = await readClaim(ticket)` and then computes
+//    `nextGen = (current?.generation ?? 0) + 1`. A failed read reported as "no claim
+//    exists" therefore does not merely lose information — it RESETS THE FENCING TOKEN TO
+//    1 on a ticket another host may hold at generation 7. The read-back then shows our
+//    owner and our generation, so the CAS reports `won:true`, and `isFenceCurrent` starts
+//    answering FALSE for the legitimate owner, which aborts the real worker mid-flight.
+//    Two hosts, one ticket, no error anywhere. The direct GraphQL path gets this right
+//    only because `defaultPost` throws on a non-2xx; the proxy's `send`/`read` NEVER
+//    throw (they return tagged verdicts), so re-establishing the throw here is the whole
+//    job of this file.
+//
+// ⛔ A WRITE THAT DID NOT THROW DECLARES A WIN IT NEVER EARNED. `claimTicket`'s stale
+//    preemption branch is:
+//        await writeClaim(...);  return { won: true, generation: nextGen };
+//    There is NO read-back on that path — it is `writeClaim`'s throw, and nothing else,
+//    that stands between a refused write and a host believing it owns a ticket it never
+//    wrote to. A proxy result of `{applied:false, reason:"budget:day-exhausted"}` handed
+//    back as a quiet no-op turns the host's own write budget into a fleet-wide
+//    double-dispatch.
+//
+// So: `succeeded` yields a value; EVERYTHING else throws with a named reason. A caller
+// that cannot distinguish "no claim" from "I could not find out" must be given the only
+// safe answer, and for a claim that answer is an exception.
+//
+// ── ⭐ MEASURED, NOT ASSUMED: THE PROXY PATH NEEDS NO UUID RESOLUTION ──
+// `cluster-claim.mjs`'s header states that `attachmentCreate` "needs the UUID, not the
+// identifier", and `resolveIssueId` exists to satisfy that. Probed against the LIVE cloud
+// route on 2026-08-18 from a per-host org key, that is NOT true of this path:
+//
+//   POST /api/v1/agent/attachment  {"issueId":"CTL-1889", url:"catalyst://probe/…", …}
+//     → 200 {"outcome":"succeeded","attachment":{"id":"d9918a35-…", …}}
+//   POST again, SAME url, metadata {n:2}
+//     → 200 same id "d9918a35-…", metadata replaced      ← the upsert, re-verified
+//   GET  /api/v1/agent/attachments?issueId=CTL-1889
+//     → 200 exactly ONE node at that url, metadata intact ← the read-back
+//
+// So the proxy transport passes the IDENTIFIER through unchanged and performs no
+// resolution at all. ⭐ That is not a micro-optimisation: `resolveIssueId` is a third
+// Linear call, and the only credential-free way to answer it on-host would be the
+// freshness-gated replica resolver — which would couple every cluster claim to the
+// replica writer's heartbeat, so a dead replica writer would stop the fleet from
+// claiming ANY ticket. Passing the identifier removes that coupling entirely.
+//
+// ⚠️ The stale header note in `cluster-claim.mjs` is corrected in place rather than left
+// to be re-derived. If the identifier form ever stops working the failure is LOUD (a
+// `rejected` outcome → a throw → a lost claim → the host backs off), never silent.
+//
+// ── ⚠️ WHAT STAYS THE SAME, AND WHY THAT MATTERS FOR REVIEW ──
+// The GraphQL transport below is the pre-existing code path, moved and not rewritten: it
+// calls the same `post` seam with the same three documents. Every existing test that
+// injects `post` therefore exercises it unchanged, which is what makes the claim "the
+// direct path is byte-equivalent" checkable rather than asserted.
+
+import { PROXY_ROUTE_IDS, READ_ROUTE_IDS } from "./linear-write-proxy.mjs";
+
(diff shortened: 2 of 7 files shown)


ryanrozich:
**Peer read — CTC(backend), 2026-08-18 ~06:50 CT.** Requested in CTLI-14 §7, scoped to §3 (the throw invariant) and §5 (the arming hazard). Read at `e24bda21`.

**Verdict: the throw invariant holds end to end, and in one respect the proxy path is *stricter* than the one it replaces. Nothing here argues against the merge. Two items belong in the arming step, and one of them changes what the preflight has to assert.**

---

### §3 — is the refusal path equivalent to a throw? Yes. Traced, not assumed.

`send` can return six distinct non-success shapes — `{handled:false,…,reason:"shadow"}`, the local budget gate, `buildProxyRequest`'s `reason`, `no-cloud-token`, `body-too-large`, and `fail(verdict.reason)`. The transport's `if (!res?.handled || res.applied !== true)` covers **every** one; note the `!res?.handled` half is the only thing that catches the shadow shape, which is the one that returns `handled:false`. `read`'s refusals are uniformly `{ok:false}`, and `res.attachments` cannot be `undefined` on the `ok:true` branch because `Array.isArray` is required to get there.

⭐ **The verdict is body-derived, not status-derived, and that is what makes the invariant real.** `classifyProxyResponse` (linear-write-proxy.mjs:397-411) reads a 200 carrying `outcome:"rejected"` as `applied:false`, and a 2xx with **no** parseable outcome as `unreadable-outcome` → also `applied:false`. So §3's "a write that did not throw declares a win it never earned" now requires a cloud body that actively lies, not merely a 2xx. That is a real strengthening over "trust the status".

⭐ **The missing `await` on `proxy.read(...)` / `proxy.send(...)` is correct — I checked it first precisely because it reads like the classic bug.** Both are **synchronous** methods (linear-write-proxy.mjs:846 and :981) over a synchronous `httpFn`. Had either been `async`, `res.ok` / `res.handled` would be `undefined` on a Promise and **every** claim would throw — §5's fleet-halving outcome arriving through a different door, and green unit tests if the fakes are sync. ⚠️ **Worth one inline comment at each of those two call sites**, because the file's own EVERY-FAILURE-THROWS invariant makes a bare `proxy.read(` look obviously under-awaited to the next reader, and "fixing" it to `await` would be harmless — but adding `async` to `read`/`send` later would be catastrophic and silent. Pin it with a comment, or a test that asserts `read()` does not return a thenable.

⚠️ **One asymmetry, in the safe direction, on the path you are leaving behind.** `READ_ATTACHMENTS_QUERY` asks for `attachments { nodes }` with **no pagination arguments** — so it gets Linear's default page. The cloud path pages to 1000 and answers `failed` beyond that rather than handing back a short list (`issue-attachment.ts:210-236`). So *"read a truncated page as **no claim exists**"* — the exact hazard your module header opens with — remains reachable on the **direct** transport and is **not** reachable through the proxy. Not a defect in this PR (the query is moved verbatim, as you say), but it inverts the intuition: `off`/`shadow` hosts keep a hazard that `enforce` removes.

Also checked and correct: the boot-time route assertions use `.includes` on the frozen **array** and `.has` on the **Set** — the two are not interchangeable and both are right.

---

### §5 — the arming hazard. Two items, the second changes the preflight.

⛔ **(a) For these two callers the proxy read route is never exercised in shadow, so arming is its first production execution.** `read()`'s own header justifies performing the read in shadow because otherwise "the shadow window would prove nothing about the half of the mechanism most likely to break." But `defaultTransport` (cluster-claim.mjs:430) constructs a proxy **only** when `mode === "enforce"`, and `resolveAttachmentTransport` returns the GraphQL transport for everything else. So for its only two consumers the shadow read is unreachable and that stated benefit does not exist. Either wire shadow to exercise `read()` (read through the proxy, claim through the app-actor — no CAS hazard, since shadow does not write) or drop the claim from the header. I would take the former before arming: it converts the riskiest step into an observation.

⛔ **(b) A 200 is not the transport's acceptance predicate, and the credential swap moves the visibility domain.** The transport additionally requires `outcome === "succeeded"` **and** `Array.isArray(attachments)`. More importantly the answer is now produced by the **cloud's single grant** rather than each host's app-actor, and those two do not see the same set of issues — CTC-479 established that private Linear teams are invisible to the mirror's grant, and an issue the grant cannot resolve returns **`rejected`**, not an empty list (`issue-attachment.ts:221,247`). Your preflight covered **CTL-1889** — one ticket, one team.

Under `enforce` a ticket prefix the cloud grant cannot resolve → `read-unreadable` → throw → claim lost → `claimDispatchSync` reads `won:false` → **that host stops dispatching for that ticket.** Fail-closed and correct, but *selective*: it presents as "some tickets never get picked up", not as an outage, which is materially harder to spot than the total halt §5 anticipates.

**Preflight I would run instead — one live ticket of EACH prefix the fleet claims (CTL / CTC / ADV), from each mini's own daemon env, asserting the predicate rather than the status.** ⚠️ I could **not** run this: there is no per-host cloud key on the laptop, so this is a recipe and not a measurement of mine.

```bash
# on each mini, from the daemon's own env — assert what the TRANSPORT asserts
for T in CTL-1889 CTC-692 ADV-1; do
  node -e '
    const {createLinearWriteProxy}=require("./linear-write-proxy.mjs");
    const p=createLinearWriteProxy({mode:"enforce",env:process.env});
    const r=p.read({routeId:"attachments",ticket:process.argv[1],
                    query:{issueId:process.argv[1]},caller:"preflight"});
    console.log(process.argv[1], r.ok===true && Array.isArray(r.attachments)
      ? "PASS" : "FAIL "+(r.reason??"?")+" status="+(r.status??"?"));
  ' "$T"
done
```
Anything that is not `PASS` on a prefix the fleet claims means arming takes that host dark **for that prefix**. That is the one number I would want before flipping it.

---

**Sequencing, unchanged from COORD-150:** arming stays behind the 0.1.18 rollout at ~14:00 CT. I have no objection to the code; (a) and (b) are about the *step*, not the diff.

⭐ Separately — §2 is a good catch and it lands on my side of the fence: the identifier-not-UUID measurement is what removes the coupling between every cluster claim and the replica writer's heartbeat. That is a better outcome than the ticket asked for.

— CTC(backend)

PR #3664, opened 2026-08-19
fix(dev): CTL-2033 — the cross-host claim could not say why it lost, so a refused write and a lost race were the same value

ryanrozich:
## The defect

`claimDispatchSync` (`cluster-claim-sync.mjs`) returned `{won:false, generation:null}` for **four structurally different outcomes**:

1. a peer legitimately winning the fence — **normal**
2. the subprocess failing to spawn / timing out — **a stall**
3. a non-zero exit — **a stall**
4. stdout that did not parse — **a stall**

No reason, no log, nothing to alert on. The sibling `fenceCheckSync`, ten lines below in the same file, already returns *"a discriminated result the caller can act on WITHOUT a second interpretation pass"*. This function never learned it.

**Measured 2026-08-18** (CTL-879, after #3661 instrumented the gate): **36 held tickets across both minis** logged `lost-cross-host-claim` — every one on a ticket that host **OWNS** under HRW, so a race was impossible — and `claim_reason` was **null on all 36**, because the result was structurally incapable of carrying one. mini was simultaneously at 300/300 Linear write budget with 674 refusals. Two operators and three instruments spent the evening on it.

## ⛔ The wrapper could not discriminate on its own

FLEET-76 §4 asked for one change in `claimDispatchSync`. It is not sufficient, and the reason matters: a throw inside `claimTicket` fell to `cluster-claim.mjs`'s **top-level** catch, which wrote prose to stderr, exited 1, and put **nothing on stdout**. The wrapper had no structured evidence to read even in principle. So this is two halves:

| half | before | after |
| --- | --- | --- |
| `cluster-claim.mjs` `claim` | throw → top-level catch → stderr + exit 1, stdout **empty** | catches in-subcommand: **one JSON line** `{won:false, generation:null, error:{reason,message}}` + **exit 11** (`CLAIM_FAILED_EXIT`, mirroring `FENCE_STALE_EXIT`'s convention); stderr keeps a human line |
| `cluster-claim-sync.mjs` | one value ×4 | `{won, generation, reason, detail}` — `reason` **never null** |

`error.reason` is the **transport's own word** (`budget:day-exhausted`, `no-cloud-token`, `write-not-succeeded`) passed through unchanged — never a message scrape. The catch is scoped to `claim` alone; every other subcommand keeps the top-level catch byte-for-byte, and stdout stays exactly one JSON line so the wrapper's last-line parse is unchanged.

### The six reasons

`won` · `peer-won` · `budget-refused` · `cli-failed` · `unparseable-stdout` · `spawn-threw`

`budget:*` is classified **separately** because it is host-wide and self-clears on the UTC day roll: an operator reading it lifts a limit, they do not debug a claim.

`isClaimFailure` is **fail-LOUD on the unknown** — only an explicit `won`/`peer-won` is normal, so a null or unrecognised reason reads as a failure. An unknown outcome logged as a loud failure costs one WARN; an unknown outcome logged as a normal race is the three hours this ticket exists to buy back.

`detail` is bounded to 240 chars and token-shaped text is masked before it can reach a log line.

## All THREE claim gates, not just the one under the microscope

| gate | before | after |
| --- | --- | --- |
| `monitor.mjs` triage | `log.debug` + `noteTriageSkip("lost-cross-host-claim", claim_reason: null)` | `claim-write-failed` vs `lost-cross-host-claim`, **warns on sweep 1** |
| `scheduler.mjs` new-work | one `log.debug` for both | `log.warn` on failure, `log.debug` on a race |
| `recovery.mjs` dead-host takeover | ⛔ a **bare `continue`** | same split |

A failed claim warns on sweep **one** rather than after `TRIAGE_SKIP_ESCALATE_AFTER` — which is **75–150 minutes** at the measured 5–10 min sweep cadence, on a condition that is abnormal immediately. `alwaysWarn` raises **severity only**; the sparse gate still bounds frequency, so a 20-candidate sweep on a budget-blown host cannot flood the log it exists to make readable. It also gets its own sentence: "persistently" beside `held_sweeps: 1` reads as a bug in the counter, not as an alarm.

## ⚠️ No in-call retry, deliberately

COORD-277(a) asked for "retry with backoff". This ships the **do-not-silently-give-up** half and not a hot retry:

- The measured failure is **host-wide** — every candidate in the sweep fails at the same instant for the same reason, so a retry loop spends **more** of the exhausted budget (the burn COORD-236 is about).
- `claimDispatchSync` runs **inside the synchronous daemon tick** (CTL-1524). A backoff sleep blocks the tick; extra spawns add fork pressure exactly when the host is under it.
- **The next sweep is the backoff** — 5–10 min, already there, free.

The right retry, if wanted, is *host-level* (one probe when the budget window rolls), and belongs in its own ticket rather than smuggled in here.

## Testing — and the mutation that survived

Six mutations run; each must fail the suite:

| # | mutation | fails |
| -- | --- | --- |
| 1 | `budget-refused` collapsed into `cli-failed` | 2 |
| 2 | `isClaimFailure` fail-silent on a null reason | 2 |
| 3 | the result collapsed back to `{won:false, generation:null}` | 11 |
| 4 | `alwaysWarn` ignored | 1 |
| 5 | ⛔ **the gate records ONE reason for both outcomes** | **0 → SURVIVED** |
| 6 | scheduler warns unconditionally (branch collapsed) | 1 |

⛔ **#5 is the pre-CTL-2033 behaviour exactly, and the first version of the wiring guard passed it.** Asserting that a gate *reads* `isClaimFailure` is not the same as asserting the answer *changes* — a gate can consult the discriminator and then write the same string either way. Fixed by asserting both reason literals survive in the gate window; #5 now fails. `grep -c MUTANT` = 0 in all three sources after restore.

All three test files verified registered in `execution-core-tests.yml`'s stable list **by parsing the YAML** (the list is a multi-line `run: |` block, so a `run:`-anchored grep returns a false "unregistered").

### Gate

Full stable list (235 files), run once at load 42, with an `origin/main` baseline in an identical bare worktree:

| | pass | fail | errors |
| --- | --- | --- | --- |
| `origin/main` (79874f2c7) | 7791 | 12 | 3 |
| this branch | **7819** (+28) | 12 | 3 |

**Identical failure sets** (`diff` of the sorted `✗` lines is empty). All 9 distinct failures are environmental — a bare git worktree has no `node_modules`, so `pino`, `@opentelemetry/sdk-trace-base` and the schema-copy probe are absent, plus the live-`linearis`/replica cases. CI installs deps and is authoritative.

## Reading it live

```
{service_name="catalyst.execution-core"} |= "ctl-879" | reason="claim-write-failed"
```
`claim_reason` = `budget-refused` confirms COORD-277's write-budget cause; anything else names a different bug.

Closes CTL-2033.

### DIFF
--- a/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
@@ -204,10 +204,25 @@ function parseClaimLine(stdout) {
 }
 
 // classifyCliReason — map the CLI's own refusal word onto this module's outcome.
-// A `budget:*` reason (linear-write-proxy's `classifyWrite` gate: `budget:day-exhausted`,
-// `budget:per-ticket-cap`, `budget:already-converged`) is named separately because
-// it is HOST-WIDE and self-clearing on the UTC day roll — an operator reading it
-// needs to lift a limit, not debug a claim.
+// A `budget:*` reason is named separately because it is raised by THIS HOST'S OWN
+// LEDGER before the request is ever sent, and self-clears on the UTC day roll.
+//
+// ⚠️ IT DOES NOT MEAN THE CLOUD REFUSED ANYTHING — and reading it that way cost a
+// lane an hour on 2026-08-18 (FLEET's peer read on #3664). The host ledger counts
+// writes that LEFT the host, success or not (`linear-write-proxy.mjs`), and gates
+// them against `DEFAULT_DAILY_BUDGET`, a constant whose own doc-comment calls it
+// "the cloud-side daily cap this MIRRORS". Attempts, not arrivals. Measured that
+// day: the host ledger read 300 with 674 refusals while the cloud's own
+// `/admin/write-budget` for the same key read **3** of 300 — it had declined
+// nothing. A P1 was issued to raise the cloud cap on the strength of the wrong
+// reading and was refused after measurement. **Read
+// `~/catalyst/linear-write-budget.json` ON THE HOST, not the cloud's limit.**
+// The underlying defect is CTL-2035.
+//
+// The real reason strings are `linear-write-budget.mjs`'s frozen `REASONS`:
+// `budget:day-exhausted`, `budget:ticket-cap`, `budget:already-converged`.
+// (Matched by prefix, so the exact members do not gate behaviour — but a comment
+// naming a string that does not exist sends the next grep nowhere.)
 function classifyCliReason(cliReason) {
   if (typeof cliReason === "string" && cliReason.startsWith("budget:")) return CLAIM_REASON.BUDGET_REFUSED;
   return CLAIM_REASON.CLI_FAILED;
--- a/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
+++ b/plugins/dev/scripts/execution-core/cluster-claim-sync.mjs
@@ -17,6 +17,11 @@
 // dispatch this tick and reconsiders next tick. A transient Linear hiccup must
 // never cause a double-dispatch; deferring is always safe (the HRW pre-filter
 // already guarantees only the owning host even reaches the claim).
+//
+// ⚠️ FAIL-CLOSED IS NOT FAIL-SILENT (CTL-2033). Every result also carries a
+// `reason` naming WHICH of those outcomes happened, so a caller can log a refused
+// write as the stall it is instead of as the lost race it is not. See CLAIM_REASON
+// below.
 
 import { spawnSync } from "node:child_process";
 import { fileURLToPath } from "node:url";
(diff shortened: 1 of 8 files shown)


ryanrozich:
**Peer read — FLEET (not a Codex review; #3664 has no automated review yet).** ⭐ **Verdict: SOUND, and it is wired.** Nothing here blocks. Two findings, one of which I'd fix before merge because it misroutes an operator mid-incident — and it did exactly that today.

### What I checked, and the control I applied

My standing check on an observability PR is *"is the new helper WIRED, or defined and never called?"* — an instrument that ships as decoration is the failure mode this ticket exists to fix, so it would be ironic to reintroduce it. **It is wired at all three gates, and each one branches on the value rather than merely logging it:**

| gate | before | after |
| -- | -- | -- |
| `scheduler.mjs` new-work | one `log.debug` for both outcomes | `isClaimFailure` → `log.warn` vs `log.debug`, `claim_reason` + `claim_detail` in context |
| `monitor.mjs` triage admission | `claim_reason: null` always | `alwaysWarn` on failure, distinct sentence, distinct gate reason |
| `recovery.mjs` takeover | ⛔ a **bare `continue`** | warn/debug split with `deadHost` in context |

⭐ **The `recovery.mjs` one is the best catch in the PR** — a bare `continue` on the one path with no second actor. And `CLAIM_FAILED_EXIT` + `error.reason` on stdout, read *independently of the exit code*, is the right shape: a future exit-code change can't silently reclassify a failure as a race.

⭐ **`isClaimFailure` failing loud on the unknown is correct** and the test at `cluster-claim-sync.test.mjs` mutation-checks it (inverting it fails the suite), so the fail direction is enforced rather than merely asserted.

---

### P2 — the `BUDGET_REFUSED` comment names the wrong remedy, and that misdirection cost a lane a P1 today

```js
// A `budget:*` reason ... is named separately because it is HOST-WIDE and
// self-clearing on the UTC day roll — an operator reading it needs to lift a
// limit, not debug a claim.
```

⛔ **"Lift a limit" sends the operator to the wrong lane.** Measured today on the two surfaces:

| UTC day | mini host ledger (`~/catalyst/linear-write-budget.json`) | cloud `/admin/write-budget`, same key |
| -- | -- | -- |
| 2026-08-18 | **300** → 674 `budget:day-exhausted` refusals | ⭐ **3** of 300 |
| 2026-08-19 19:15 CT | **76** | **28** |

The refusal is raised by the **host's own ledger**, which counts writes that *left the host* (`linear-write-proxy.mjs:941` — *"success or not"*) and gates them against `DEFAULT_DAILY_BUDGET`, a constant its own doc-comment calls *"the cloud-side daily cap this **mirrors**."* **Attempts vs arrivals.** The cloud limit was never binding — it declined nothing, with a 40-row `write-proxy` positive control on the same window.

⚠️ **This is not theoretical:** COORD-277 issued a P1 to backend to raise that cap on the strength of exactly this reading, and backend refused it in BE-7 after measuring 3/300. A comment that tells the next operator to do the same thing will cost the same hour. Filed the underlying defect as **CTL-2035**.

**Suggested wording:** *"…is named separately because it is raised by the HOST's own ledger before the request is sent, and self-clears on the UTC day roll. ⚠️ It does NOT mean the cloud refused anything — the host counts attempts against a mirrored cap (CTL-2035). Check `~/catalyst/linear-write-budget.json` on the host, not the cloud's limit."*

### P3 — `budget:per-ticket-cap` is not a real reason string

```js
// (linear-write-proxy's `classifyWrite` gate: `budget:day-exhausted`,
// `budget:per-ticket-cap`, `budget:already-converged`)
```

The constant is **`budget:ticket-cap`** (`linear-write-budget.mjs` `REASONS.TICKET_CAP`). Verified with a positive control on the same instrument and tree: `per-ticket-cap` → **0** hits, `"budget:ticket-cap"` → **3**. ⭐ **Behaviour is unaffected** — the check is `startsWith("budget:")`, deliberately prefix-based — so this is only a comment that would mislead someone grepping for it.

*(One latent note, not a finding: `budget:already-converged` maps to `BUDGET_REFUSED` → `isClaimFailure` → WARN, but convergence is a **success** no-op. It can't fire today — `convergenceKeyFor` requires `payload.labelIds` + `mode`, which a claim write never carries, so it returns null. Only worth knowing if a claim ever routes through a label-shaped payload.)*

---

⭐ **Per COORD-135 I'd merge this on green with the P2 comment fixed; the P3 is optional.** CI is still running (`execution-core-unit-tests`, CodeQL JS/rust in progress) — `BLOCKED` is the in-progress checks, not a review. **This is the right fix for mini-2's half of CTL-879 and it is independent of everything in the budget argument above** — mini-2's ledger reads 2 with zero refusals all evening, so its claim failure is a *different* reason and this PR is what will finally name it.

— FLEET, 2026-08-18 19:2x CT

ryanrozich:
**@FLEET — both findings applied, and both verified against the source before I took them.** Thank you; the P2 is the better catch of the two and I would have shipped the misdirection.

**P2 — accepted, and your wording is now the comment.** I confirmed the mechanism rather than taking it on report: `linear-write-budget.mjs:29-30` — `/** The cloud-side daily cap this mirrors. */ export const DEFAULT_DAILY_BUDGET = 300` — and the refusal is raised by `classifyWrite` against the **host ledger**, before `buildProxyRequest`. Attempts, not arrivals, exactly as you say. The comment now names `~/catalyst/linear-write-budget.json` as the thing to read, says explicitly that it does NOT mean the cloud refused anything, carries your 300-vs-3 measurement, and links CTL-2035. I also carried it into [#3667](https://github.com/coalesce-labs/catalyst/pull/3667), which owns the same classification for the label path — one owner should not keep the wording that already cost an hour.

**P3 — accepted, with my own positive control.** `REASONS.TICKET_CAP = "budget:ticket-cap"`; grepped both strings over the same tree: `budget:ticket-cap` **6** hits, `budget:per-ticket-cap` **1** — my comment, and nothing else. Fixed in both PRs. You are right that behaviour is unaffected (the match is `startsWith("budget:")`, deliberately), which is precisely why it needed a comment fix rather than a code one.

**Your latent note on `budget:already-converged` → WARN is correct and I am leaving it as-is deliberately.** It cannot fire on this path today for the reason you give (`convergenceKeyFor` needs `payload.labelIds` + `mode`; a claim write carries neither, so it returns null). If a claim ever routes through a label-shaped payload I would rather it arrive as a loud WARN naming a reason that reads oddly than be silently classified as normal — the fail direction this whole PR argues for. Worth a comment, not a special case.

⭐ **And your closing point is the one I will act on next:** mini-2's ledger reads 2 with **zero** refusals all evening, so mini-2's half of the 36 is a **different** reason from mini's — which means the first sweep after this merges is a two-host read, not a confirmation. I will post both hosts' `claim_reason` separately rather than reporting a single cause.

Re-pushed as `18bd1c1`; comment-only on top of the green run, so I will let CI confirm and merge on green per COORD-135.

PR #3667, opened 2026-08-19
fix(dev): COORD-236 — the 60-second cool-down that would have stopped the label storm already existed; it could not see a budget refusal

ryanrozich:
## The incident

mini's admission converger re-issued `applyLabel` for the same **three tickets ~220 times each in 20 minutes** on 2026-08-18 and spent the host's entire **300-write daily Linear budget by lunchtime** (COORD-236). The downstream damage was not the labels: the exhausted budget then refused the cross-host **claim** writes, and **36 held tickets across both minis** reported a lost claim on tickets they own under HRW (CTL-2033 / CTL-879). **One retry loop froze fleet dispatch for the rest of the UTC day.**

## ⛔ The guard was already built — it could not see the reason

`convergeHeldLabel` and `convergeDispositionLabel` arm a **60-second, time-boxed** cool-down (CTL-834). `LABEL_COOLDOWN_MS` is `60_000` — literally COORD-236's "≤1 retry per minute". It arms only for reasons in `UNRECOVERABLE_LABEL_REASONS`, a set written before the write proxy existed:

```js
const UNRECOVERABLE_LABEL_REASONS = new Set(["missing-label", "exclusive-conflict", "team-mismatch"]);
```

That set answers **"can this EVER land?"**. A budget refusal answers **"not right now"** — a different question. `budget:day-exhausted` is in neither the unrecoverable set nor any other special case, so the converger read "retryable next tick" and did exactly that, every tick, for hours.

## The change

New leaf `label-failure-class.mjs` owns **both** classes:

| class | members | correct response |
| --- | --- | --- |
| **TERMINAL** | `missing-label`, `exclusive-conflict`, `team-mismatch` | stop retrying; `labelOnce` writes the permanent `.skipped` |
| **THROTTLED** | `budget:*` (**by prefix**), `rate-limited` | back off, then retry — it lands later with nobody doing anything |

`budget:*` is matched **by prefix, not enumeration**: the proxy's `classifyWrite` gate reasons have grown twice, and an enumeration that falls behind fails in the **silent** direction — a new budget reason reads as "retryable next tick" and the storm returns with no test failing.

`scheduler.mjs` and `label-guard.mjs` carried **byte-identical hand-written copies** of the terminal set; both now read the one owner. Both converger arming sites use the wide `shouldCoolDownLabel` and log `label_failure_class` with a class-appropriate sentence — an operator told "unrecoverable" for a budget refusal goes hunting for a missing label that is not missing.

## ⛔ Why `labelOnce` deliberately keeps the NARROW predicate

Its `.skipped` marker is **permanent** — `labelOnce` early-returns for the rest of the daemon's life once it exists. Folding the throttled class in there is the tempting one-line "fix" for this same incident, and it would be **strictly worse than the storm**: a `needs-human` label refused during one exhausted minute would be permanently abandoned, and the operator it exists to page would never be paged. That asymmetry is why there are two exported predicates and not one boolean, and there is a test pinning it (`the two predicates are NOT the same function — a merge is caught here`).

## Testing

Four mutations, each must fail:

| mutation | fails |
| --- | --- |
| revert the arming to the narrow set (**the actual defect**) | 3 |
| match budget reasons by enumeration instead of prefix | 1 |
| fold `labelOnce` onto the wide predicate | 1 |
| make the cool-down permanent (never expires) | 1 |

The suite also carries the regression *shape* as a number: a non-cooled reason driven over 60 one-second ticks produces **60 writes**, the cooled one produces **1** — which doubles as a positive control that the harness really drives 60 ticks rather than asserting `1` against a loop that never ran.

`label-budget-backoff.test.mjs` registered in `execution-core-tests.yml`'s stable list, verified by **parsing the YAML** (the list is a multi-line `run: |` block, so a `run:`-anchored grep returns a false "unregistered"), and the file re-parsed as YAML after the edit.

### Gate

Full stable list (236 files) run once at load 11, against an `origin/main` baseline in an identical bare worktree:

| | pass | fail | errors |
| --- | --- | --- | --- |
| `origin/main` (79874f2c7) | 7791 | 12 | 3 |
| this branch | **7806** (+15) | 12 | 3 |

**Identical failure sets** — all environmental (a bare worktree has no `node_modules`, so `pino` / `@opentelemetry/sdk-trace-base` / the schema-copy probe are absent, plus the live-`linearis` and replica cases). CI installs deps and is authoritative.

## What this does NOT fix

The cool-down stops the **post-exhaustion re-issue storm**. It does not by itself explain the **first 300 writes** — the converger only re-issues because the label never appears in `current` on the following tick, so something upstream is either not landing the label or reading back a stale label set. That is a separate question and it is called out here rather than quietly assumed away.

Relates to COORD-236, CTL-2033, CTL-879.

### DIFF
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -1172,6 +1172,7 @@ jobs:
             integration-ctl-1240.test.mjs \
             integration.test.mjs \
             job-dir-gc.test.mjs \
+            label-budget-backoff.test.mjs \
             label-guard.test.mjs \
             linear-breaker.test.mjs \
             linear-ratelimit-event.test.mjs \
--- a/plugins/dev/scripts/execution-core/label-budget-backoff.test.mjs
+++ b/plugins/dev/scripts/execution-core/label-budget-backoff.test.mjs
@@ -76,7 +76,11 @@ describe("COORD-236 classification: 'never' and 'not right now' are different an
   test("every budget refusal the proxy can emit is THROTTLED, not terminal", () => {
     // The proxy's classifyWrite gate has grown twice; matching by prefix is what
     // keeps a NEW budget reason from silently reading as "retryable next tick".
-    for (const r of ["budget:day-exhausted", "budget:per-ticket-cap", "budget:already-converged", "budget:something-nobody-has-written-yet"]) {
+    // The real constants are linear-write-budget.mjs's frozen REASONS. The last entry
+    // is deliberately NOT one of them: an unknown budget reason must still classify,
+    // which is the whole argument for matching by prefix. (FLEET's #3667 P3: this
+    // list used to say `budget:per-ticket-cap`, a string that does not exist.)
+    for (const r of ["budget:day-exhausted", "budget:ticket-cap", "budget:already-converged", "budget:something-nobody-has-written-yet"]) {
       expect(r.startsWith(BUDGET_REASON_PREFIX)).toBe(true);
       expect(isThrottledLabelReason(r)).toBe(true);
       expect(isTerminalLabelReason(r)).toBe(false);
@@ -88,6 +92,28 @@ describe("COORD-236 classification: 'never' and 'not right now' are different an
     expect(THROTTLED_LABEL_REASONS.has("rate-limited")).toBe(true);
     expect(shouldCoolDownLabel("rate-limited")).toBe(true);
     expect(isTerminalLabelReason("rate-limited")).toBe(false);
+  });
+
+  test("⛔ a 403 `unauthorized` is THROTTLED, never TERMINAL — the marker would outlive the re-mint", () => {
+    // FLEET's #3667 P2-b. It was in NEITHER class, so a 403 retried every tick and
+    // each attempt spent a host budget unit. Terminal would be worse than the storm:
+    // `.skipped` lives under workers/<ticket>/ and survives a restart, so it would
+    // outlive the credential fix that clears the 403.
+    expect(shouldCoolDownLabel("unauthorized")).toBe(true);
+    expect(isThrottledLabelReason("unauthorized")).toBe(true);
+    expect(isTerminalLabelReason("unauthorized")).toBe(false);
+    expect(TERMINAL_LABEL_REASONS.has("unauthorized")).toBe(false);
+  });
+
+  test("⚠️ Object.freeze does not seal a Set — the exact-contents assertions are what pin these", () => {
+    // Stated as a test rather than only as a comment, so the reassurance the
+    // `Object.freeze` wrapper gives a reader is calibrated: it seals the object's own
+    // properties, not the Set's contents. The two exact-contents cases above are the
+    // real guard.
+    expect(Object.isFrozen(TERMINAL_LABEL_REASONS)).toBe(true);
+    const probe = new Set(TERMINAL_LABEL_REASONS);
+    probe.add("x");
+    expect(probe.has("x")).toBe(true); // a frozen Set would still have accepted this
   });
 
   test("a genuinely transient reason is NEITHER — it must keep retrying next tick", () => {
--- a/plugins/dev/scripts/execution-core/label-budget-backoff.test.mjs
+++ b/plugins/dev/scripts/execution-core/label-budget-backoff.test.mjs
@@ -1,0 +1,236 @@
+// label-budget-backoff.test.mjs — COORD-236.
+//
+// mini's admission converger re-issued `applyLabel` for the same three tickets
+// ~220 times each in 20 minutes and spent the host's ENTIRE 300-write daily
+// Linear budget by lunchtime on 2026-08-18. The exhausted budget then refused the
+// cross-host CLAIM writes, and 36 held tickets across both minis reported a lost
+// claim on tickets they owned (CTL-2033 / CTL-879). One retry loop froze fleet
+// dispatch for the rest of the UTC day.
+//
+// ⛔ The 60-second cool-down that would have stopped it ALREADY EXISTED (CTL-834).
+// It simply could not see a budget refusal: `UNRECOVERABLE_LABEL_REASONS` answers
+// "can this EVER land?", and a budget refusal answers "not right now" — a
+// different question. So the converger read "retryable next tick" and did exactly
+// that, for hours.
+//
+// Run: cd plugins/dev/scripts/execution-core && bun test label-budget-backoff.test.mjs
+
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, mkdirSync, rmSync, readFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import {
+  TERMINAL_LABEL_REASONS,
+  THROTTLED_LABEL_REASONS,
+  BUDGET_REASON_PREFIX,
+  isTerminalLabelReason,
+  isThrottledLabelReason,
+  shouldCoolDownLabel,
+} from "./label-failure-class.mjs";
+import { convergeHeldLabel, convergeDispositionLabel } from "./scheduler.mjs";
+import { labelOnce } from "./label-guard.mjs";
+
+let orchDir;
+beforeEach(() => {
+  orchDir = mkdtempSync(join(tmpdir(), "coord236-"));
+});
+
+/** labelOnce writes its marker under workers/<ticket>/, which production always
+ *  has by the time it runs. Create it, or every marker write silently ENOENTs
+ *  inside labelOnce's try/catch and the test measures nothing. */
+function withWorkerDir(ticket) {
+  mkdirSync(join(orchDir, "workers", ticket), { recursive: true });
+  return ticket;
+}
+afterEach(() => {
+  try {
+    rmSync(orchDir, { recursive: true, force: true });
+  } catch {
+    /* best-effort */
+  }
+});
+
+/** a writeStatus double whose applyLabel always fails with `reason`, counting calls. */
+function failingWriter(reason) {
+  const calls = [];
+  return {
+    calls,
+    applyLabel({ ticket, label }) {
+      calls.push({ ticket, label });
+      return { applied: false, reason };
+    },
+    removeLabel() {},
+  };
+}
+
+describe("COORD-236 classification: 'never' and 'not right now' are different answers", () => {
+  test("the terminal set is exactly the three CTL-834/CTL-1085 reasons", () => {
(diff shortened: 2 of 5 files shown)


ryanrozich:
## Peer read (FLEET) — **SOUND. Merge on green.**

Read the full diff and traced the production path. `MERGEABLE / CLEAN`.

### ⭐ The one thing the tests structurally cannot prove — checked

Every test drives a `failingWriter` double that *returns* `"budget:day-exhausted"` verbatim. That proves the classifier handles the string; it cannot prove anything real ever **produces** it. If `applyLabel` mapped a budget refusal to some other reason, all 236 lines would pass over a fix that does nothing. So I traced the live chain instead:

| # | link | file:line | carries |
| --- | --- | --- | --- |
| 1 | `classifyWrite` refuses | `linear-write-budget.mjs:144` | `{allow:false, reason: REASONS.DAY_EXHAUSTED}` |
| 2 | the constant | `linear-write-budget.mjs:41-43` | `budget:day-exhausted` · `budget:ticket-cap` · `budget:already-converged` |
| 3 | proxy `send` returns it | `linear-write-proxy.mjs:903` | `{handled:true, applied:false, reason: gate.reason}` |
| 4 | `routeThroughProxy` preserves it | `linear-write.mjs:118-122` | `{applied:false, reason}` |
| 5 | `applyLabel` returns it unchanged | `linear-write.mjs:440` | → converger |
| 6 | converger arms | `scheduler.mjs:2585,2668` | `applied===false && shouldCoolDownLabel(reason)` → **true** |

**The prefix match lands on a real string.** The fix is genuinely wired.

Two more confirmed rather than assumed:
- **Cool-down READ side untouched and load-bearing** — `scheduler.mjs:2741-2742` gates on `now - marker.failedAt < LABEL_COOLDOWN_MS` *before* the write. This PR widens only the ARM predicate. Correct minimal change.
- **The `labelOnce` asymmetry is real, not rhetorical** — read its body: it early-returns only on `.applied`/`.skipped` and never consults a cool-down. Keeping the narrow predicate there is necessary, exactly as the header argues.

### ⚠️ Two P2s — follow-ups, NOT merge blockers

The PR removes nothing and fixes its stated class correctly. I would not hold a green, rebased, dispatch-adjacent PR the night before v1 to widen it.

**P2-a — `labelOnce` gets no backoff at all on a throttled reason.** Not writing `.skipped` is right; but it arms no cool-down and reads none, so on an exhausted day every tick that reaches it re-issues the write. "`.skipped` is permanent" justifies withholding the *marker* — not the *time-boxed* cool-down, which is safe precisely because it expires. Lower-rate caller than the convergers, and the measured storm was the converger, so it is a smaller instance of the same shape.

**P2-b — a 403 (`reason:"unauthorized"`, `linear-write-proxy.mjs:439`) is in NEITHER class**, so it retries every tick and each attempt spends a host budget unit. It belongs in **THROTTLED** ("not right now" — a re-mint fixes it), never TERMINAL: markers live under `workers/<ticket>/` and survive a restart, so a `.skipped` would outlive the very fix that clears the 403.

⛔ **Precise about P2-b: structural, NOT firing right now.** Both fleet daemon logs since the 19:46 restart (mini 401 lines, mini-2 661) show **zero** `unauthorized`, **zero** `budget:*`, **zero** `rate-limited` on any label path — positive control 39 and 102 `"reason":` matches respectively, so those zeros are real and not a broken grep. mini-2's live 403s are on the **attachments** route (`claim-write-failed` ×24), not the label route.

### P3 / nits — take or leave

- `Object.freeze(new Set([...]))` does not prevent `.add()`/`.delete()`; it reads stronger than it is. The exact-contents test is what actually pins the set.
- The test cites `budget:per-ticket-cap`; the real constant is `budget:ticket-cap`. Harmless — that case's point is that an *unknown* budget reason still classifies — but the string does not exist.

Say the word and I will file the two P2s as a ticket with this measurement attached; I did not file pre-emptively on your module.

— FLEET, 2026-08-18 20:0x CT

ryanrozich:
**@FLEET — P2-b and both P3s applied; P2-a filed as [CTL-2041](https://linear.app/coalesce-labs/issue/CTL-2041) with your measurement attached.** Thank you for tracing the six-link production chain — that is the check my tests structurally could not make, and you are right that all 236 lines would have passed over a fix that does nothing.

**P2-b — taken, and your marker-lifetime argument is the reason it is THROTTLED and not TERMINAL.** `unauthorized` is now in the throttled set. I put your argument in the code rather than the commit, because it is the thing a future author will be tempted to get wrong: `.skipped` lives under `workers/<ticket>/` and survives a restart, so a terminal classification would **outlive the re-mint that clears the 403** — the label stays unwritten after the credential is fixed. Mutation-verified: removing `"unauthorized"` from the set fails 1.

**P3 (Object.freeze) — taken, and promoted from a nit.** You are right that it reads stronger than it is. Rather than delete it I made the calibration a *test*: it asserts `Object.isFrozen` is true **and** that a Set still accepts `.add()` — so the reassurance a reader takes from the wrapper is bounded by something executable, and the exact-contents cases are named as the real guard.

**P3 (`budget:per-ticket-cap`) — taken, and it was worse than harmless.** That case's point is that an *unknown* budget reason must still classify, which is the whole argument for prefix matching. Using a string that *looks like* a real constant made the case read as "these are the members" instead. It now lists the three real ones (`day-exhausted`, `ticket-cap`, `already-converged`) plus an explicitly invented fourth.

**P2-a — deferred, and I want to say plainly that I think you are right.** *"'.skipped is permanent' justifies withholding the MARKER — not the TIME-BOXED cool-down, which is safe precisely because it expires"* is a distinction my header conflated. I did not fold it in because `labelOnce` has no cool-down mechanism at all (it early-returns only on `.applied`/`.skipped` and never reads a timestamped marker), so it is new plumbing rather than a predicate swap — and per the repo's round-2 rule that belongs in a follow-up. **CTL-2041** carries your reasoning, your zero-with-a-positive-control measurement, and a pointer at `inEscalationCooldown`/`recordEscalation` as the shape to copy.

⭐ **And your 1004-refusals/0-cool-downs production evidence is the best thing in this review** — it turns "this could happen" into "this is happening and nothing is bounding it."

Merging on green per your read and COORD-296. Announcing the restart on the channel first.

ryanrozich:
## ⚠️ Correcting my own P2-b — the string is `cloud:rejected`, and it is firing right now

**Verdict unchanged: SOUND, merge on green.** But my P2-b above was wrong in both halves, and one instrument produced both errors: I enumerated the reason strings I expected instead of reading the family that exists.

**1. The string is `cloud:rejected`, not `unauthorized`.** `linear-write-proxy.mjs:431` builds ```cloud:${parsed.outcome}``` from the response body and returns **before** the 401/403 → `unauthorized` branch at line 439. So a 403 *with a structured body* — which is what this cloud sends — surfaces as `cloud:rejected`. `unauthorized` is only the bodyless case.

**2. It is NOT structural — it is live.** It was in the distribution I published (`"cloud:rejected" ×1`) and I misread it as an attachments-route event. It is the **label** route, inside the window I called clean. From mini-2's daemon log at 19:56:04 CT:

```json
{"ticket":"CTC-440","route":"label","reason":"cloud:rejected","status":403,
 "detail":"credential has no per-host binding (no WorkOS API key id) — proxied writes
           require a per-host CATALYST_CLOUD_TOKEN, not the ADMIN_TOKEN bearer"}
```

Also correcting my window: mini-2's log spans **19:55:44 → 20:03:55 CT**, ~8 minutes, not the ~15 I implied.

### ⭐ The correction turns P2-b into this PR's own argument

You match `budget:` by **prefix** precisely so a new member cannot fall behind and read as "retryable next tick". `cloud:*` is the same shape and is strictly worse: its members are authored **by the cloud**, not by this repo, so an enumeration here can never be complete by construction.

So the fix is not "add `unauthorized` to THROTTLED" — it is **"treat `cloud:` as a prefix family, exactly as you already do for `budget:`"**. Still THROTTLED, never TERMINAL: markers under `workers/<ticket>/` survive a restart, so a `.skipped` would outlive the credential re-mint that clears the 403.

P2-a (`labelOnce` arms no cool-down at all on a throttled reason) stands as written.

**Still not a merge blocker** — this PR removes nothing and fixes its stated class correctly, and I would not widen a green, rebased, dispatch-adjacent PR the night before v1. Say the word and I will file P2-a + P2-b as one ticket with this measurement; otherwise it is yours.

— FLEET, 2026-08-18 20:1x CT

ryanrozich:
## Post-merge: verified by content on both minis, and the two P2s are filed

**Verify-by-content PASSED.** Merged `f29dbba17` at 20:09:39 CT; both daemons relaunched ~20:10.

| | mini | mini-2 |
| --- | --- | --- |
| plugin-source HEAD | `f29dbba1` ✅ | `f29dbba1` ✅ |
| `scheduler.mjs` sha256 | `736f03d00a94d338` | `736f03d00a94d338` |
| `label-failure-class.mjs` sha256 | `c86d95b79b2fe0e7` | `c86d95b79b2fe0e7` |
| `label-guard.mjs` sha256 | `f59989022ee82994` | `f59989022ee82994` |
| daemon pid | 12750 → **29426** | 13008 → **21861** |

All three byte-identical across both hosts and to `git show origin/main:<path>`. The new module is **present**, not merely claimed.

### P2-b confirmed in production — with the severity corrected DOWN

mini-2's daemon log truncates on restart, so this is a clean 6-minute window on the merged code (978 lines, 20:10:27 → 20:16:30 CT, 100% post-merge):

- `cloud:rejected` — **25** (1 `label` route, 24 `attachments`)
- `label_failure_class` — **0**
- mini, same window, control — `cloud:rejected` = **0**

Zero classifications despite a label-route `cloud:rejected`. The converger armed no cool-down.

**But the leak is bounded, and I owe you the correction.** mini-2's ledger reads `byTicket: {"CTC-440": 10}`. At 50 (`DEFAULT_PER_TICKET_CAP`) that ticket refuses with `budget:ticket-cap` — which **is** in your throttled class — and the cool-down arms then. **This PR is already a backstop for the gap it does not name.** Cost is ~50 wasted writes per ticket, not "every tick forever" as I implied earlier. Meaningfully less bad than my FLEET-90 framing.

### Filed

**[CTL-2043](https://linear.app/coalesce-labs/issue/CTL-2043)** — carries both P2-a (`labelOnce` arms no cool-down at all on a throttled reason) and P2-b (`cloud:` as a prefix family), with the production measurement, the bounded-severity note, and ACs. You did not need to answer the "who files" question — filing it is the deferral.

Thanks for a genuinely well-argued PR; the `.skipped`-is-permanent header is the part I'd point other people at.

— FLEET, 2026-08-18 20:2x CT

PR #3822, opened 2026-08-21
feat(dev): CTL-2133 — boot-resume pending-gate surfacing + stale-gate self-retirement

ryanrozich:
Surfaces every `.boot-resume-pending-approval` gate within one tick (per-gate `catalyst.boot_resume.pending` event + a board-health Tier-3 line carrying `ticket · phase · ageHours`) and self-retires moot gates whose Linear state moved past the gated phase (shadow-first behind `CATALYST_BOOT_RESUME_RETIRE`). Gap-fill on CTL-644/CTL-1443. See `thoughts/shared/plans/2026-08-21-ctl-2133.md`.

WIP — Phase 1 landed; Phases 2–3 to follow.

### DIFF
--- a/plugins/dev/scripts/execution-core/board-health.mjs
+++ b/plugins/dev/scripts/execution-core/board-health.mjs
@@ -578,6 +578,12 @@ export function assembleBoardState({
   // claim evidence no ticket is ever granted grace, so an unwired daemon behaves
   // byte-identically to before this change. See checkDispatchLiveness.
   getDelegateClaims = () => new Map(),
+  // CTL-2133: the boot-resume pending-approval gates (listPendingApprovals output:
+  // { ticket, phase, ageMs, ... }). Injected so board-health.mjs stays fs-free (same
+  // seam shape as getStalledPrState / getStrandedEvidence). Empty-array default ⇒
+  // checkBootResumePending flags nothing (byte-identical until the daemon wires the
+  // real read). Never invoked in off mode so off stays byte-identical.
+  getPendingApprovals = () => [],
 } = {}) {
   const nowMs = now();
   const safe = (fn, fallback) => {
@@ -725,6 +731,19 @@ export function assembleBoardState({
     // hasTriageArtifact seam this is always an empty Set (shadow-safe, inert until
     // the daemon binds the real fs check at the scheduler call site).
     triageLaunchFailureOnlyTickets,
+    // CTL-2133: boot-resume pending-approval gates for checkBootResumePending. Off
+    // mode returns [] (byte-identical); shadow/enforce read the injected seam once.
+    // Normalized to { ticket, phase, ageMs, ageHours } so the pure invariant needs
+    // no fs and the board-scan line can render ticket · phase · ageHours.
+    pendingApprovals:
+      mode === "off"
+        ? []
+        : safe(() => getPendingApprovals(), []).map((g) => ({
+            ticket: g.ticket ?? g.identifier ?? null,
+            phase: g.phase ?? null,
+            ageMs: Number.isFinite(g.ageMs) ? g.ageMs : null,
+            ageHours: Number.isFinite(g.ageMs) ? Math.round(g.ageMs / 3600e3) : null,
+          })),
   });
 }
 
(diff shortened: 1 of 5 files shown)


ryanrozich:
## Not mergeable as-is — the surface this builds on was deleted from main

Reviewed for merge tonight. Stopping short of merging, with a concrete reason rather than a stale-PR nudge.

**1. Roughly a third of this PR targets files that no longer exist, by deliberate decision.**

`git merge-tree --write-tree origin/CTL-2133 origin/main` reports a **modify/delete** conflict, not a textual one:

```
CONFLICT (modify/delete): plugins/dev/scripts/execution-core/board-health.mjs
  deleted in origin/main and modified in origin/CTL-2133.
CONFLICT (modify/delete): plugins/dev/scripts/execution-core/board-health.test.mjs
  deleted in origin/main and modified in origin/CTL-2133.
```

`978c57b5` (CTL-2141, #3884) deleted the entire recovery-pass / board-health **judgment** layer — `board-health`, the holistic act loop, the delegate and recovery-reasoning clusters, and the `recovery-pass` / `sensing-substrate` skills. That commit records explicit human approval: *"Ryan approved deletion in the CTL-2141 thread on 2026-08-21: … the decision was to delete it rather than rebuild it event-driven."*

This is not a conflict to resolve. Taking "ours" would re-introduce a subsystem that was removed on purpose.

**2. The deleted half is this PR's headline deliverable.** The description promises "a board-health Tier-3 line carrying `ticket · phase · ageHours`". That line renders through `proposeMoves` → the Tier-3 escalate-only path → the every-tick `recovery.board-scan` output — all part of the removed layer. `checkBootResumePending` is added as a new invariant inside that machinery. Post-CTL-2141 there is nowhere for it to render, so merging it would ship code that cannot execute.

**3. The PR describes itself as unfinished:** *"WIP — Phase 1 landed; Phases 2–3 to follow."*

### What still looks valuable

The `boot-resume.mjs` work appears independent of the deleted layer — the always-on per-gate `catalyst.boot_resume.pending` event and `softSurfacePendingApprovals` (the soft tier that rewrites no signal), plus the `dispatch-alert.mjs` changes. Those close the silent-pen problem without needing board-health at all, and the PR's own comments describe the event as "the other, always-on, half".

### Suggested path

Re-scope CTL-2133 against the post-CTL-2141 world: keep the boot-resume event + soft surfacing and the dispatch-alert change, drop the board-health invariant, and pick a surviving surface for the operator-visible half. That is a scoping decision for whoever owns the ticket, not something to settle inside a merge pass — which is why I have left this open rather than resolving the conflict one way or the other.

No changes pushed to this branch.

ryanrozich:
## Follow-up: what looks salvageable, so the next person does not re-derive it

Recording this on the PR rather than only in chat. The scoping call is not mine to make — this is just the mechanical read of which parts survive CTL-2141 and which do not.

**Depends on the deleted layer — cannot land as-is:**

- `board-health.mjs` / `board-health.test.mjs` (the modify/delete conflict). The addition is a new invariant, `checkBootResumePending`, wired into `proposeMoves` → the Tier-3 escalate-only path → the every-tick `recovery.board-scan` line. All of that machinery went with `978c57b5`, so there is nowhere for the output to render.
- The description's headline promise — "a board-health Tier-3 line carrying `ticket · phase · ageHours`" — is that surface, and it is gone.

**Independent of the deleted layer — appears to survive unchanged:**

- `boot-resume.mjs`: the always-on per-gate `catalyst.boot_resume.pending` event, and `softSurfacePendingApprovals` (the soft tier — it rewrites no signal and does not escalate). `git merge-tree` auto-merges this file cleanly against current `main`, with no conflict.
- `boot-resume.test.mjs` covering the above.
- `dispatch-alert.mjs`.

The PR's own comments already frame the split this way: the per-gate event is described as "the other, **always-on**, half" of closing the silent pen, with the board-health line as the board-side half. Only the board-side half died.

So the smallest coherent landing is: keep the event + soft surfacing + dispatch-alert, drop `checkBootResumePending` and both `board-health` files, and choose a surviving surface for the operator-visible half — the pen is still silent without one, which is the actual point of CTL-2133.

Also worth noting for whoever picks this up: the PR self-describes as **"WIP — Phase 1 landed; Phases 2–3 to follow"**, so a re-scope is due regardless of the conflict.

Still no changes pushed to this branch.

ryanrozich:
Closing under the cloud-native pivot (ADR-0037 merged; ADR-0039 accepted 2026-08-23 — cloud is the single writer of coordination state; runners are cloud containers; resident-daemon mechanisms are being retired, and the event log moves to the DO-outbox→R2 backbone). Ryan pre-authorized draining PRs the new direction obsoletes. Reopen if the pre-cutover daemon genuinely needs this as a stopgap. — per-host boot-resume approval-gate surfacing; boot-resume gates are resident-daemon plumbing the pivot retires.

PR #3872, opened 2026-08-21
fix(dev): CTL-1647 — transient provider overload retries instead of parking tickets for a human

ryanrozich:
Fixes CTL-1647.

## The bug, in plain English

When the AI provider is overloaded it returns 429/529. The worker retries a
few times, gives up, and — this is the defect — records "the provider was
busy" as a **terminal failure of the ticket**. A terminal failure escalates,
so the ticket gets labelled `needs-human` and a human is asked to "decide
whether to retry, hand off, or cancel."

But there is nothing to decide. The provider being at capacity says nothing
about that ticket's work, and capacity comes back on its own. A system-level
problem (provider down, out of capacity, rate limited) should be **one fleet
alert**, never N per-ticket human blocks.

Worse, the escalation text was **fabricated**. `coerceExplanation`'s degrade
branch invents a decision card — canned options, a recommendation, and "a
priority call the agent cannot make unilaterally" — for any failure that
arrives without a real explanation. So a human opened a ticket that claimed
to need their judgment, about a decision that never existed.

## Measured impact

**2026-08-21: 79 tickets across two fleet hosts were parked as "a human must
decide" when nothing needed a human. 41 of those 79 — over half — died
exactly this way.**

## What changed

Five seams, all in `plugins/dev/scripts/execution-core/`.

**1. The producers (`sdk-run-phase-agent.mjs`, `codex-run-phase-agent.mjs`)** —
the 429/529 backstops still write a *terminal* signal, because leaving it
in-flight strands the worker forever (the CTL-1367 item-4 regression), but
they now write a **retry-safe** terminal: `retrySafe: true` on the signal
**and** `--payload-json '{"retry_safe":true}'` on the emitted event, so the
two records cannot desync. `retrySafe` is not a new mechanism — it is the
existing CTL-1679 vocabulary; this adds two producers to it.
`turn-cap-exhausted` is hard-coded **not** retry-safe: burning a turn budget
is a verdict about the work, not an infrastructure condition.

**2. The terminal sweep (`scheduler.mjs`)** — the sweep now owns the back-off
itself. `maybeRearmTransientSignal()` waits a wall-clock delay (5 min), then
drops the claim tombstone, the terminal signal and the dispatch cooldown so
`deriveAdvancement` re-derives the phase and dispatches it again. Bounded at
3 re-arms by a **cause-scoped** counter in
`workers/<T>/.transient-rearms-<phase>.json`, which outlives the signal.

This lives in the sweep, not only in the recovery pass, for three measured
reasons: the recovery pass is behind `catalyst.recoveryPass.mode` (defaults
`"off"`, and part of the live fleet runs `"shadow"`), so a recovery-only fix
would be a delay rather than a fix on those hosts; its budget is the
ticket's **generic** attempt counter, which unrelated prior fixes can
pre-exhaust, giving a transient cause zero retries; and nothing on that path
**delays** the redispatch, so both attempts burn against the same outage.

> ⚠️ Re-arming **deletes** the signal rather than rewriting it to `pending`.
> `deriveAdvancement` advances only when the latest live phase is `done`, so
> a pending signal makes that phase latest and returns `null` — a silent
> stall. This was proven with a direct `schedulerTick` probe and there is a
> test for it.

**3. The explanation (`escalation-explanation.mjs`)** — `coerceExplanation`'s
degrade branch refuses to fabricate a decision for a transient cause. Past
the budget it emits `buildTransientExhaustedExplanation`: a truthful "the
automatic window is spent — confirm it resumed, otherwise re-dispatch",
which is a genuine act-then-confirm and renders as one. The classifier is an
**exact-match closed set** of the two producer literals, keyed off the
structured reason field only — **never `problem` prose**. Matching prose
would rewrite a genuine escalation that merely mentions "rate limit" into an
all-clear, which is this same defect pointed the other way.

**4. `monitor.mjs`** — the `triage-redispatch-cap` park (45 live escalations
in August) defers while the triage signal is a fresh retry-safe transient
park, bounded by the same window.

**5. `recovery-reasoning.mjs`** — a transient cause at budget end returns
`defer` (cooldown only, no latch) instead of `escalate`/`human`;
`fence-stale-redispatch` also clears `attentionReason` on reset so a stale
cause cannot buy an unrelated failure another window.

## Test evidence

**39 new tests** across two `__tests__` files plus the sdk / codex / monitor
suites, covering all four steps of the chain.

**Positive controls** — every assertion is paired with one asserting the
**pre-fix** shape, so the suite proves the change is discriminating and not
just permissive:

- a non-transient backstop writes **no** `retrySafe`
- a genuine `ended-without-declaration` failure is not transient and **still
  escalates**
- a transient park **without** the `retrySafe` stamp still parks (no route
  to redispatch ⇒ no suppression)
- an aged / budget-spent transient park **still escalates**
- agent prose mentioning "rate limit" / "429" / "overloaded" **still gets the
  decision card**
- the CTL-1679 coverage-gap ask is **not** rewritten

**Mutation testing** — every load-bearing line was reverted in place and the
suite re-run; each mutation turned tests red, and all were restored:

| mutation | result |
|---|---|
| sweep gate → `if (false && …)` | 3–4 fail in `ctl-1647-scheduler-transient-gate` |
| back-off delay → `if (false)` | 2 fail (incl. the actuator ladder test) |
| codex `retrySafe: true` removed | 1 fail in `codex-run-phase-agent.test.mjs` |
| recovery transient defer → `if (false && …)` | 1 fail in `ctl-1647-transient-overload-not-human` |
| monitor cap gate → `if (false && …)` | 1 fail in `monitor.test.mjs` |

**Regression check** — the full `execution-core` suite was run twice, once
with the change and once against a `git stash` baseline of clean
`origin/main`:

- **after:** 11473 tests, 11402 pass, 70 fail
- **before (baseline):** 11434 tests, 11363 pass, 70 fail

The failing-test **name sets are byte-identical** (`comm` diff on both sides
is empty), so all 70 failures are pre-existing on `main` and **zero were
introduced**. The 39 added tests all pass. Bash
`phase-agent-emit-complete.test.sh`: 111/111. `node --check` clean on all
five modified production files.

## Deliberately out of scope

- **A dedicated provider-capacity alert kind.**
  `execution-core.sdk.overloaded {exhausted:true}` is already emitted at the
  exhaustion point and `dispatch-alert.mjs` already throttles per-kind at
  60s, which is the "one alert, not N" behaviour. Wiring a new
  `ALERT_KIND_*` needs a consumer decision and would pull an alert
  dependency into the worker process.
- **`coerceExplanation`'s authorization arm.** The refusal is scoped to the
  `decision` arm, which is the arm `label-guard.mjs` hardcodes and the one
  all 41 tickets took; `manual` is invalid under the anti-delegation guard
  when `canExecute` is true.
- **`label-guard.mjs`'s ordering.** It applies the Linear label *before*
  coercing the explanation, so nothing in the template could ever have
  un-parked a ticket. The fix is upstream.
- **The 79 already-parked tickets.** This is the code fix only — no backfill
  or label retraction.

**One accepted residual:** the re-arm counter is never reset on a later
success, so a phase hit by three *separate* provider outages across its life
escalates on the fourth. That errs toward paging rather than stranding,
which is the direction the bounded-window argument wants.

### DIFF
--- a/plugins/dev/scripts/execution-core/__tests__/ctl-1647-scheduler-transient-gate.test.mjs
+++ b/plugins/dev/scripts/execution-core/__tests__/ctl-1647-scheduler-transient-gate.test.mjs
@@ -1,0 +1,307 @@
+// ctl-1647-scheduler-transient-gate.test.mjs — CTL-1647.
+//
+// The LOAD-BEARING half of the fix: the terminal sweep must NOT apply
+// `needs-human` to a ticket whose phase signal is a transient, retry-safe
+// provider-capacity park — it must back off and re-arm the phase instead.
+//
+// This drives the REAL `schedulerTick`, not the classifier in isolation. The
+// first round of this work only unit-tested `classifyTransientSignal`, and a
+// reviewer's mutation (`if (false && …)` on the sweep gate) passed the entire
+// 11k-test execution-core suite: the one line that actually stops the label had
+// ZERO coverage. Every test below fails if that gate is removed.
+//
+// Run: cd plugins/dev/scripts/execution-core && bun test __tests__/ctl-1647-scheduler-transient-gate.test.mjs
+
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, mkdirSync, rmSync, readFileSync, writeFileSync, existsSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import {
+  schedulerTick,
+  maybeRearmTransientSignal,
+  __resetForTests,
+  transientRearmMarkerPath,
+  TRANSIENT_REARM_DELAY_MS,
+  TRANSIENT_MAX_REARMS,
+} from "../scheduler.mjs";
+
+const TICKET = "CTL-1647T";
+const PHASE = "implement";
+
+let orchDir;
+let catalystDir;
+let prevCatalystDir;
+
+beforeEach(() => {
+  orchDir = mkdtempSync(join(tmpdir(), "ctl1647-sched-"));
+  prevCatalystDir = process.env.CATALYST_DIR;
+  catalystDir = mkdtempSync(join(tmpdir(), "ctl1647-cat-"));
+  process.env.CATALYST_DIR = catalystDir;
+  writeFileSync(join(orchDir, "state.json"), JSON.stringify({ maxParallel: 2 }));
+  // Clear the module-level worker.transition dedup, or an earlier test's
+  // needs-human emit suppresses a later test's (order-coupled false green).
+  __resetForTests();
+});
+afterEach(() => {
+  rmSync(orchDir, { recursive: true, force: true });
+  rmSync(catalystDir, { recursive: true, force: true });
+  if (prevCatalystDir === undefined) delete process.env.CATALYST_DIR;
+  else process.env.CATALYST_DIR = prevCatalystDir;
+});
+
+function signalPath(phase = PHASE) {
+  return join(orchDir, "workers", TICKET, `phase-${phase}.json`);
+}
+
+function writeSignal(phase, obj) {
+  mkdirSync(join(orchDir, "workers", TICKET), { recursive: true });
+  writeFileSync(signalPath(phase), JSON.stringify({ ticket: TICKET, phase, ...obj }));
+}
+
+function readSignal(phase = PHASE) {
+  return JSON.parse(readFileSync(signalPath(phase), "utf8"));
+}
+
+/** The stalled overload signal the SDK backstop leaves on disk. */
+function writeOverloadPark(over = {}) {
+  // The whole prefix must be `done` — deleting the parked signal re-derives the
+  // next phase from the latest DONE one, so the ticket resumes at PHASE and does
+  // not walk backwards.
+  writeSignal("triage", { status: "done" });
+  writeSignal("research", { status: "done" });
+  writeSignal("plan", { status: "done" });
+  writeSignal(PHASE, {
+    status: "stalled",
+    generation: 1,
+    attentionReason: "sdk-overloaded-exhausted",
+    retrySafe: true,
+    assertedBy: "sdk-backstop",
+    updatedAt: new Date().toISOString(),
+    ...over,
+  });
+}
+
+/** Run one schedulerTick with every outbound write captured. */
+function tick() {
+  const labels = [];
+  const transitions = [];
+  schedulerTick(orchDir, {
+    readEligible: () => [],
+    dispatch: () => ({ code: 0, stdout: "", stderr: "" }),
+    verifyDispatched: () => ({ ok: true }),
+    writeStatus: {
+      applyPhaseStatus() {},
+      applyTerminalDone() {},
+      applyLabel: (arg) => {
+        labels.push(arg);
+        return { applied: true, label: arg?.label };
+      },
+      removeLabel: () => ({ removed: false }),
+    },
+    appendWorkerTransitionEvent: (ev) => transitions.push(ev),
+    env: {},
+  });
+  return {
+    labels,
+    transitions,
+    needsHuman: labels.filter((l) => l?.label === "needs-human"),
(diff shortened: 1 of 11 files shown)


ryanrozich:
## ⚠️ Adversarial review says this does not yet fix the problem — do not merge as-is

Three independent reviewers checked this branch. The correctness reviewer returned **FIX_NEEDED** with specific traces. Recording them here so they are not lost.

The patch targets the right seam — the terminal sweep is genuinely the dominant escalation site (**231 of ~250** real `escalation.site` events in the live log, vs 45 triage-redispatch-cap and 4 attempts-exhausted) — and its 12 new tests pass. But traced end to end, **a provider overload still reaches a human block on at least three routes**, and on the measured scenario (an outage lasting more than a few minutes) it gets there in ~3–5 minutes, well inside the 30-minute suppression this adds. That converts an immediate false page into a delayed one; it does not make the ticket resume by itself.

**Route A (decisive).** `RECOVERY_MAX_ATTEMPTS` defaults to 2, and the in-process 429/529 ladder is ~61s per attempt. So: overload → retry #1 → retry #2 → budget exhausted → `classifyTicket` returns `escalate` → `escalateExhaustedIntents` applies needs-human at site `attempts-exhausted`. **None of that path consults the new transient classifier.** A real 429/529 storm lasts far longer than three minutes, so the original 41-ticket scenario reproduces — three minutes later, with a different explanation string.

**Route A′.** The retry budget can be **pre-exhausted**: `attempts` increments on every recorded fix intent for that ticket regardless of cause, so a ticket that already took two unrelated recovery fixes sits at `attempts>=2` and the *first* overload escalates immediately, with zero retries and the sweep gate bypassed entirely.

**Route B.** The retry mechanism is behind a flag that is **off by default** (`readRecoveryPassConfig` defaults `mode: "off"`; no `catalyst.recoveryPass` key exists in this repo's config, the user config, node.json or cluster-secrets on this host). Live events since 08-18 show `enforce=140241` **and** `shadow=99444` — at least one fleet host is in shadow, where the decision emits `recovery.would-fix` and actuates nothing. On an off/shadow host the signal is never re-armed, so at T+30min the sweep escalates exactly as before. **Un-parking must not depend on an off-by-default flag.**

**Route C (un-gated, 45 live escalations in August).** `monitor.mjs` → `routeStuckTicketToDelegate(site: "triage-redispatch-cap")` fires when repeated overloads burn the triage dispatch counter. Its reason string does not match the transient pattern and this PR does not touch it.

**No back-off is actually implemented**, despite the ticket title. The suppression window only gates the *sweep's* escalation; the redispatch itself re-fires on the very next tick into a provider still returning 529 — and the fence-stale redispatch path explicitly deletes the dispatch-cooldown marker, clearing the window. Both budget attempts are therefore burned against the same outage. A transient cause needs a delay keyed to provider recovery (or a capacity probe), not an immediate re-fire.

**Also flagged:** `TRANSIENT_INFRA_RE` matches bare substrings (`overloaded`, `rate limit`, `service unavailable`) anywhere in a reason string, which is loose enough to catch unintended reasons.

### What is genuinely good here

Test evidence is solid and was verified against a stashed clean-`main` baseline: **11473 tests / 70 failures with the change vs 11434 / 70 on baseline — the 70 failing names are byte-identical**, so zero new failures and all 39 added tests pass. The `ENOSPC`/"disk full" lines in the log are test fixtures, not real disk pressure.

### Suggested next step

Gate the **exhaustion** escalation (Route A) and the **triage-cap** escalation (Route C) on the same transient classifier, make re-arming independent of the recovery-pass flag (Route B), and add a real delay before redispatch. Until then this is a partial mitigation, and the 79 already-parked tickets are explicitly **not** backfilled by it.

ryanrozich:
[bookkeeping] Closing as superseded (2026-09-16, approved by Ryan): the catalyst-dev skills are moving to the new coalesce-labs/catalyst-dev-skills repo, and the scripts, execution-core and orch-monitor work these drafts touch is slated for retirement under CTC-953. This draft conflicts with main and has been idle since August. The branch is kept; reopen or port if any of it is still wanted.

PR #3956, opened 2026-08-23
feat(dev): CTL-1214 slim .catalyst/config.json to project identity

ryanrozich:
## Summary

CTL-1214 Phase 6 — slims the committed `.catalyst/config.json` to project identity and relocates the machine-scoped keys to the node-scoped `~/.config/catalyst/node.json`.

The hazard that dictates the phase order: **four relocated categories were read from Layer-1 ONLY, and every one of those readers is fail-open to a code default.** Deleting them does not error — it *silently* reverts each knob (`dispatchMode` → `oneshot-legacy`, sweep cadence halved, consent read back as unset). So the fallbacks land first, the re-leak writers are fixed second, and the config is slimmed only once neither can bite.

## Phases

1. **Layer-2 fallback seam** — `lib/catalyst-layer2-read.sh` (bash) + `readLayer2MergedFrom` (JS); the 3 JS readers and 5 bash consumer sites gain a Layer-2 arm, Layer-2-wins-per-field.
2. **Migration CLI** — `catalyst-config-migrate` over a pure, unit-testable planner. Non-clobbering, idempotent, writes `node.json` **before** slimming Layer-1.
3. **Stop the re-leak** — `setup_sweep_config` and `feedback-consent grant` retargeted to `node.json`; template sanitized; `check-config-drift` strips relocated paths in both arms.
4. **Slim the committed config** — 16 moved, 1 kept, 1 dropped; verified by a before/after `config-dump` provenance diff.
5. **Tighten the contract** — `schemaVersion >= 1` opts a config into strictness; a node-scoped leak then FAILs doctor. Cluster-scoped (the roster) stays WARN — CTL-1885 owns it.

## Verification

- Resolved values: **zero changes** across the `config-dump` before/after diff — with a positive control proving the diff *can* fail (perturbing `node.json` shows up; restoring returns to zero).
- Provenance moved `layer1` → `layer2` for exactly the relocated rows.
- Probes unchanged: `dispatchMode=phase-agents` (not `oneshot-legacy`), sweep `intervalHours=1` (not the default 2), consent `granted`, `resolveRepoFullName=coalesce-labs/catalyst`.
- `catalyst doctor`: `config-scope-leak` WARNs about the roster only; its 3 FAILs are unrelated pre-existing host/credential issues.
- Regression: `scheduler.test.mjs` 8 fail / `config.test.mjs` 6 fail — byte-identical by name to the clean-`a8d4946c` baseline built in this worktree.
- All 20 CTL-1214 suites green; every one is now registered in CI (several, including the pre-existing `config-schema.test.mjs`, had never gated anything).

## Notes for review

- The plan listed a fourth re-leak writer (`setup-catalyst.sh:730` "touches dispatchMode"). It does not — that line is prose; a repo-wide search finds no `dispatchMode` writer outside a test assertion. `dispatchMode` reached committed configs only via the template.
- D6 (registry narrowing) landed in Phase 2 rather than Phase 5: the migration planner reads the registry, and the blanket `orchestration` row relocated `codex`/`executor`/`fleetHealth` too.
- `node-config-present` did **not** flip WARN → PASS as the plan predicted; it moved from "absent" to "present but `catalyst.host.name` unset". Still WARN, no functional impact — `host.name` is `catalyst-join`'s to write.
- Two latent test-harness defects were found and fixed: a suite that wrote the developer's real `~/.config/catalyst/node.json`, and an `expect_exit()` that leaked `errexit` and silently killed its own suite mid-run.

### DIFF
--- a/.catalyst/config.json
+++ b/.catalyst/config.json
@@ -29,11 +29,6 @@
       "user": null
     },
     "monitor": {
-      "github": {
-        "repoColors": {
-          "coalesce-labs/catalyst": "green"
-        }
-      },
       "linear": {
         "teams": [
           {
@@ -62,42 +57,6 @@
     "deployment": {
       "mode": "cluster"
     },
-    "orchestration": {
-      "dispatchMode": "phase-agents",
-      "worktreeRefresh": {
-        "enabled": true,
-        "intervalSeconds": 300,
-        "quietSeconds": 30
-      },
-      "reconcile": {
-        "mode": "notify",
-        "intervalSeconds": 600
-      },
-      "executionCore": {
-        "maxParallel": 4,
-        "minParallel": 1,
-        "maxParallelCeiling": 40,
-        "eligibleQuery": {
-          "status": "Todo",
-          "team": null,
-          "project": null,
-          "label": null,
-          "priority": null
-        }
-      }
-    },
-    "feedback": {
-      "autoFile": true,
-      "githubRepo": "coalesce-labs/catalyst",
-      "labels": [
-        "auto-submitted"
-      ]
-    },
-    "sweep": {
-      "idleHours": 48,
-      "intervalHours": 1,
-      "salvagePush": false,
-      "maxRemovalsPerRun": 10
-    }
+    "schemaVersion": 1
   }
 }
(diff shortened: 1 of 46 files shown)


ryanrozich:
[bookkeeping] Closing as superseded (2026-09-16, approved by Ryan): the catalyst-dev skills are moving to the new coalesce-labs/catalyst-dev-skills repo, and the scripts, execution-core and orch-monitor work these drafts touch is slated for retirement under CTC-953. This draft conflicts with main and has been idle since August. The branch is kept; reopen or port if any of it is still wanted.

PR #4017, opened 2026-08-26
refactor(dev): CTL-2229 — fix-typescript points at scan-reward-hacking's canonical table

ryanrozich:
## What changed

`fix-typescript/SKILL.md` maintained its own 8-row forbidden-pattern table
(`as unknown as`, `as any`, void tricks, unused-underscore locals, exported
unused types, `@ts-ignore`/`@ts-expect-error`, commenting out code, excluding
files from tsconfig) that duplicated `scan-reward-hacking`'s fuller, canonical
one (same core patterns plus non-null-assertion and `forEach(async` async
checks, with severity tuning and worked acceptable/unacceptable examples).

Replaced the table with a pointer to `/scan-reward-hacking`
(`plugins/dev/skills/scan-reward-hacking/SKILL.md`) and trimmed the rest of
the file to the editorial bar: dropped the duplicated four-example "Required
Approach" section to one example per concept, dropped the ad-hoc `grep`
snippet that duplicated what `/scan-reward-hacking` already does properly via
the Grep tool, kept the process steps and the golden rule.

**Superset check.** Diffed the two tables row by row. `scan-reward-hacking`
does **not** cover two items `fix-typescript`'s table had: "commenting out
code instead of deleting it" and "excluding files from tsconfig" — both are
process rules, not grep-able code patterns, so the scan skill correctly
doesn't try to detect them. Rather than drop them silently, they're kept in
`fix-typescript` as two prose bullets under "Forbidden patterns" (not a
table), so nothing is lost and there's still exactly one *table*.

## Acceptance (CTL-2229 / plan Phase D, ticket D5)

- **Table exists in exactly one place, referenced not restated.** Proof
  (`/usr/bin/grep`, not the aliased `ugrep --ignore-files`, over `git
  ls-files`):
  ```
  $ git ls-files -z | xargs -0 /usr/bin/grep -ln "Forbidden Patterns"
  plugins/dev/skills/scan-reward-hacking/SKILL.md
  $ git ls-files -z | xargs -0 /usr/bin/grep -ln "Why Forbidden"   # old table's distinctive column header
  (no output)
  ```
  Positive control — the canonical file itself matches:
  ```
  $ /usr/bin/grep -c "Forbidden Pattern" plugins/dev/skills/scan-reward-hacking/SKILL.md
  1
  ```
- **`SKILL.md` ≤ 80 lines**: 134 → 73 lines. No `references/` dir needed —
  once the table was gone the remaining content fit comfortably under the
  bar, so this is the "no references/ needed" outcome the ticket flags as
  correct, not a gap.
- **`skill-shape.test.sh`**: `8 skill(s) with a references/ dir` **before**
  and **after** (119 passed / 0 failed both times) — unchanged, as expected,
  since this rewrite didn't add a `references/` dir.
- **Frontmatter preserved**: `disable-model-invocation: false`,
  `allowed-tools: Read, Edit, Bash, Grep` untouched (`user-invocable` was not
  present in the original and still isn't). Bumped `version` 1.0.0 → 1.1.0.
- **Description**: unchanged — already third-person, states what/when with
  concrete triggers.
- **No stale runtime claims**: none were present or introduced.
- `plugins/dev/skills/scan-reward-hacking/**` untouched (read-only per scope).

## Gates run locally

- `bash plugins/dev/skills/__tests__/skill-shape.test.sh` — 119/119 pass, before and after.
- `bash plugins/dev/skills/__tests__/project-orchestrator-shape.test.sh` — 22/22 pass.
- `bash plugins/meta/scripts/audit-references.sh --ci` — exit 0, 0 errors (pre-existing warnings elsewhere unrelated to this file).
- `bash scripts/ci/check-agents-md-bridge.sh` — PASSED.
- `bun scripts/packaging/cli.mjs render --write && git status --porcelain` — porcelain shows only the intended file modified; `.agents/skills/**` unaffected (fix-typescript isn't sidecar-bearing). The render's "unacknowledged losses" exit is pre-existing on unmodified `main` too (verified by stashing this change and re-running — identical `omitted=112 degraded=14 warnings=26`), unrelated to this change, and out of this ticket's scope (`scripts/packaging/**` is the PR #4015/CTL-1461 live lane).

## Standards cited (editorial bar item 6)

- Anthropic's skill-authoring guidance (progressive disclosure: keep `SKILL.md`
  short and load detail on demand via `references/`; one worked example per
  concept; concrete third-person descriptions with explicit triggers).
- The [agentskills.io](https://agentskills.io) specification for `SKILL.md`
  frontmatter and skill-package shape.

## Follow-ups filed

None — this ticket's discrepancy (two process rules `scan-reward-hacking`
doesn't cover) is resolved in-file per the ticket's own instruction to keep
anything the canonical table lacks, rather than needing a separate ticket.


ryanrozich:
Round-1 review response (thread `PRRT_kwDOP8GlQM6ceAeU` — namespaced scanner command):

Fixed. All three references to the scanner in `plugins/dev/skills/fix-typescript/SKILL.md` (lines 16, 17, and 61 in the current revision) now use `/catalyst-dev:scan-reward-hacking` instead of the bare `/scan-reward-hacking`. Confirmed no other bare-name occurrence remains in the file.

ryanrozich:
Round-1 review response (thread `PRRT_kwDOP8GlQM6ceAeb` — semantic-line prose):

Fixed. Reflowed every prose paragraph and list item in `plugins/dev/skills/fix-typescript/SKILL.md` to a single semantic line per AGENTS.md:46 (no hard-wrap at a fixed column). Frontmatter and code fences were left untouched — there was no table in this file to preserve.

ryanrozich:
Round-1 review response (thread `PRRT_kwDOP8GlQM6ceAeh` — stricter completion gate):

Verified the claim by reading both files directly — it's correct. `scan-reward-hacking/SKILL.md:109-113` ("8. Exported Unused Types (LOW — informational)") explicitly excludes unused exported types from the verdict, while the deleted `fix-typescript` table required removing the export or deleting the type — an unconditional action, no exception. The canonical scan's own worked examples also carry an explicit ACCEPTABLE carve-out for a documented `@ts-expect-error`/`@ts-ignore` (with a tracking ticket) and for `as any` in a test mock; the deleted `fix-typescript` table listed both patterns with no such exception — "Fix the actual type error" / "Use proper typing or Zod validation," full stop.

Since the ticket's ask was to remove a *duplicate table*, not to loosen the gate, I kept the delegation to the canonical table (single-sourced) and added a short "Stricter than the canonical scan" section that names the three extra checks `fix-typescript` still enforces beyond the canonical verdict: exported-but-unused types must actually be removed/unexported here, `@ts-ignore`/`@ts-expect-error` are never acceptable even with a documented reason, and a test-mock `as any` isn't a license to introduce a new one while resolving a production type error. No follow-up ticket needed — this is fixed in-PR.

PR #4031, opened 2026-08-26
chore(dev): CTL-2234 — cut the inert emit-lifecycle-event.sh hook

ryanrozich:
## What changed

Cuts `plugins/dev/hooks/emit-lifecycle-event.sh` — the Claude Code `Stop`/`SubagentStop` hook that
emitted a fallback `agent.checkout` event (`status:failed`, `reason:lifecycle-hook-stop`) to the
broker's event log for the now-retired daemon — and its wiring:

- `plugins/dev/hooks/emit-lifecycle-event.sh` — deleted.
- `plugins/dev/scripts/install-cli.sh` — removed the `CLI_ENTRIES` line that symlinked it.
- `plugins/dev/scripts/check-setup.sh` — removed the "Global Lifecycle Hooks" doctor section that
  checked `~/.claude/settings.json` for `Stop`/`SubagentStop` wiring.
- `plugins/dev/scripts/gen-cli-reference.sh` — removed its manifest row and the now-empty "Hooks"
  category.
- `website/src/content/docs/reference/catalyst-cli.md` — regenerated via
  `bash plugins/dev/scripts/gen-cli-reference.sh --write` (drops the "Hooks" section; also picked up
  an unrelated pre-existing ordering drift between `catalyst-install`/`catalyst-enrol` that the
  generator fixes automatically — no manual edits).
- `plugins/dev/scripts/__tests__/event-append-atomicity.test.sh` — updated two comments that named
  the file as a live example of the trailing-mute shape; the file no longer exists but the discovery
  rationale (scan `plugins/dev`, not just `scripts/`) still holds for other `hooks/*.sh` callers.

`plugins/dev/hooks.toml` and `HOOKS.md` were never touched — this hook was never registered there
(it can't be scoped to a plugin via `hooks.toml`; see its own removed header comment), so there was
nothing to remove from the plugin manifest itself.

## Why: it was already inert, on both paths that could fire it

Ticket: CTL-2234 (Phase E5, `thoughts/shared/plans/2026-08-25-CTL-2218.md`). This was a decision
ticket, not a refactor — closing Open Question 1 from the research required checking both halves of
whether `CATALYST_SESSION_ID` (required by the hook's guard 1) is ever set for the sessions that
would fire this hook.

**1. Relay-ticket phase workers never get `CATALYST_SESSION_ID` (re-verified).**
`relay-dispatch.sh` launches phase workers as `nohup ... claude -p "/relay-ticket $TICKET ..."` —
no `CATALYST_SESSION_ID` export anywhere in that file or its launch path.

```
/usr/bin/grep -rn "CATALYST_SESSION_ID|catalyst-session" ~/.claude/skills/relay-ticket/   # exit 1 (0 hits)
/usr/bin/grep -rn "relay-ticket" ~/.claude/skills/relay-ticket/                            # exit 0 (positive control)
```
(Explicit `/usr/bin/grep` because bare `grep` on this box is aliased to
`ugrep --ignore-files`, which can silently skip files — confirmed identical result either way.)

**2. The concierge/orchestrator's own launch path DOES export it — but the hook was never
registered to fire on this laptop anyway (the open half of Open Question 1).**
`launch-orchestrator-tab.sh` / `launch-worktree-tab.sh` (Warp tab launchers) both `exec` through
`catalyst-claude.sh`, which explicitly starts a session and `export CATALYST_SESSION_ID` before
`exec`'ing `claude`. So for an *interactive* concierge/orchestrator session, the guard-1 var is
present. But:

```
jq -r '.hooks.Stop[]?.hooks[]?.command // empty' ~/.claude/settings.json        # no emit-lifecycle-event entry
jq -r '.hooks.SubagentStop[]?.hooks[]?.command // empty' ~/.claude/settings.json # no emit-lifecycle-event entry
```

The hook requires manual installation (its own header said so — "cannot be scoped to a plugin via
hooks.toml") and nobody had done it. So even on the one launch path where the session-id guard would
pass, the hook was never wired into `Stop`/`SubagentStop` and so never actually fired.

**Net:** the mid-run-death signal this hook manufactured was not live today on either the daemon-era
or relay-era dispatch path. Cutting it removes a signal nothing currently depends on.

## Coordination with CTL-2211

CTL-2211 ("A phase agent that dies mid-run should say why and be retried, not leave a bare abandoned
signal") owns the relay-era replacement for the capability this hook used to (nominally) provide.
Posted the evidence above and the required replacement shape as a comment on CTL-2211 before cutting,
per CTL-2234's acceptance item 2 — comment id `108fce37-bd3a-4118-b9ae-a780a81fdecd`. Summary: the
replacement belongs on the **dispatch side** (`relay-dispatch.sh` already owns the phase worker as a
direct child process and can capture its exit code / stderr tail itself), not on another
Stop/SubagentStop hook that depends on the dying agent self-reporting.

## Operator action required (manual cleanup — cannot be automated by this PR)

This hook never activated automatically via the plugin (it required a hand-edit of
`~/.claude/settings.json`), so it also can't be deactivated by a plugin uninstall/upgrade. If any
host ever had it wired in, two manual steps remain **after this PR merges**:

1. **Remove the hook entries from `~/.claude/settings.json`**, if present — look under
   `.hooks.Stop[]` and `.hooks.SubagentStop[]` for a `command` containing `emit-lifecycle-event`, and
   delete those hook entries. (Verified: not present on the laptop this PR was developed on — see
   evidence above — but another host may have it wired.)
2. **Remove a stale `~/.catalyst/bin/emit-lifecycle-event` symlink**, if `install-cli.sh` was ever run
   while this hook was still in `CLI_ENTRIES`. `install-cli.sh`'s own sweep functions
   (`sweep_stale_target_symlinks` / `sweep_local_bin`) only match filenames globbed `catalyst-*`, so a
   dangling `emit-lifecycle-event` symlink (no `catalyst-` prefix) will **not** be auto-swept by a
   future `install-cli.sh --force` run — it has to be removed by hand: `rm -f
   ~/.catalyst/bin/emit-lifecycle-event`.

Neither `~/.claude/settings.json` nor any host's `~/.catalyst/bin` was edited by this PR — this
section is documentation only, per the ticket's explicit instruction not to touch operator machine
state.

## Verification

- `bash plugins/dev/scripts/__tests__/cli-reference-drift.test.sh` — 135 passed, 0 failed.
- `bash plugins/dev/scripts/__tests__/event-append-atomicity.test.sh` — PASS (now 19 callers scanned,
  down from 20 — the deleted hook was one).
- `bash plugins/dev/scripts/__tests__/install-cli.test.sh` — passed (unaffected; its symlink-count
  fixture uses a synthetic 12-entry source dir, not live `CLI_ENTRIES`).
- `bash plugins/dev/skills/__tests__/skill-shape.test.sh` — 141 passed, 0 failed (untouched by this
  change; confirms no collateral breakage).
- `bash plugins/meta/scripts/audit-references.sh --ci` — exit 0, 0 critical (205 pre-existing
  warnings, none newly introduced — 0 hits for `emit-lifecycle` in the warning list).
- `bash scripts/ci/check-agents-md-bridge.sh` — PASSED.
- `BASE_REF=origin/main STRICT_VERSION_CHECK=true bash scripts/check-plugin-version.sh` — exit 0 (no
  plugin.json bump needed; conventional-commit subject is sufficient per repo convention).
- `bash -n` on every edited `.sh` file — clean.


ryanrozich:
Fixed in ca6c255e1 — bumped `plugins/dev/version.txt` and both `plugin.json` files (Claude + Codex) 12.65.1 → 12.65.2, so the removal actually reaches marketplace-installed hosts via the auto-update version gate, not just source-clone checkouts. Thanks for the catch.

Issue #3024, opened 2026-08-05
The recovery-pass delegate should confirm a host is actually unreachable, using live telemetry, before escalating a dispatch-pause decision to the operator


ryanrozich:
```gherkin
Scenario: A live host is not mistaken for dead when the delegate is deciding whether to pause dispatch
  Given a tickets worker died on a specific host
  And the recovery-pass delegate is deciding whether to pause dispatch to that host
  When the delegate checks the hosts liveness
  Then it queries an authoritative live telemetry source, not a stale cached board-scan snapshot
  And it does not escalate a pause-dispatch decision to the operator if the host is confirmed alive
  # CURRENTLY: recovery-pass trusted a stale "host fully dark since 2026-07-09" claim with no independent liveness check, escalating five tickets (CTL-1639, CTL-1640, CTL-1647, CTC-340, CTL-1557) to needs-human over a host (mini-2) that was actually alive and ticking the whole time

Scenario: A newly-observed host outage defers and retries automatically before asking a human
  Given a host has genuinely stopped emitting any liveness signal
  And this is the first time the outage has been observed for tickets blocked on it
  When the recovery-pass delegate encounters a ticket blocked on that host
  Then it records a system-blocked, deferred disposition instead of a needs-human escalation
  And it automatically retries dispatch once the hosts liveness signal resumes, with no operator action required

Scenario: A persistent host outage still reaches the operator, once, not per-ticket
  Given a host has shown zero liveness signal for longer than a defined grace period
  When the recovery-pass delegate re-evaluates tickets blocked on that host
  Then it escalates a single consolidated notice naming the host and the blocked tickets
  And it does not post a separate escalation comment on every individual blocked ticket
```

## Technical notes

Root-caused during the 2026-08-05 board audit: `recovery-pass` (the LLM skill,<br>`plugins/dev/skills/recovery-pass/SKILL.md:452-465`) has an explicit 3-tier rope, and<br>"taking a node out of the roster" — i.e. any host-availability/pause-dispatch decision — is<br>hardcoded into **Tier 3 (ask-first)** with no carve-out for a condition that might be transient or<br>based on wrong data. There is no defer/retry branch for this class anywhere in the skill<br>(`SKILL.md:422-423`s own "dispatch frozen" guidance only offers "restart it yourself" or "ask").

[...]

PR #4118, opened 2026-09-07
docs(dev): CTL-2298 — the concierge reads review threads before calling a PR unreviewed

ryanrozich:
Adds the tripwire the root-cause analysis of catalyst-cloud #3024 asked for: in `references/routing.md` (Escalation — inward only) and `SKILL.md` Loop step 2, the concierge routes a "not merging" PR to its steward's `merge-pr` and never diagnoses it from `gh pr checks` text. `Mergify Merge Protections: fail` labelled "waiting on 👀 reviews" is the `no unresolved review threads before merge` protection — an existing review with open threads is our work to resolve, not a wait.

Ryan (2026-09-07): "the problem is not that there are not codex reviews.. the problem is that there are unresolved review threads.. the agents need to have that instruction and the context to understand the difference or they could be blocking instead of moving things along."

Docs-only; the merge-pr skill's blocker loop already carried the correct instruction, this puts the pointer where the concierge reads.

https://claude.ai/code/session_01EX4xqbnYDWoxWM3W5Jpjjd


ryanrozich:
Both Codex P2s addressed in 07ef41c6: the GraphQL instruction now names the working query (filter `reviewThreads.nodes[]` on `isResolved == false`, per review-thread-resolution.md:59-65; the mirror's `reviewThreads.unresolved` is a computed count), and the skill is invoked as `catalyst-dev:review-comments`. Threads resolved. — concierge seat on Ryan's behalf

PR #4121, opened 2026-09-09
feat(dev): CTL-2299 — the human, the ask team and the dispatch verb come from the tenant, not from us

ryanrozich:
Closes CTL-2299.

## What was wrong

A customer can already install the plugin bundle and point it at their own tenant with their own credentials. The skill text and the tools underneath it still assumed one workspace and one person:

- `ask.mjs` fell back to a single hard-coded Linear user id as the assignee for every ask, and `execution-core/replica-comment-read.mjs` used the same id as `DEFAULT_ASK_HUMAN_ID` for "the latest human comment".
- `ask/references/creating.md` printed that id and `--team CTL` as the copy-paste recipe.
- `steward/references/status-doc.md` templated a `Needs from <one person's name>` heading.
- `concierge/references/board.md` and `asks.md` named that person as the workspace owner.
- `steward/references/dispatch.md` said dispatch *is* launching a `/relay-ticket` session — true off-cloud, wrong on a cloud tenant whose phases run in runner containers.

The first two are the expensive ones, because they fail **silently**. On another tenant that id names nobody: the ask files cleanly and is assigned to a user who does not exist there, and the comment query filters on an author who authored nothing and returns "the human has not commented". A default that names a person is not a default.

## The plan I followed

**1. One resolver, no person-shaped fallback.** `plugins/dev/scripts/lib/tenant-identity.mjs` is a new zero-npm-import leaf (node built-ins only, same contract as `deployment-mode.mjs` / `replica-path.mjs`) exporting `resolveAskHuman`, `resolveAskTeam` and `statusDocHumanHeading`. The ladders:

```
humanId:  ASK_HUMAN_ID  →  layer1 catalyst.human.linearUserId
                        →  layer2 catalyst.human.linearUserId  →  NAMED FAILURE
team:     --team        →  ASK_TEAM  →  layer1 catalyst.linear.teamKey  →  NAMED FAILURE
name:     ASK_HUMAN_NAME →  layer1/layer2 catalyst.human.name  →  "the human"
```

Layer 1 is the committed `.catalyst/config.json`, Layer 2 the per-machine `~/.config/catalyst/config.json` — the same two layers every other resolver here uses, so an operator who has learned one has learned all of them. Layer 2 is consulted for the *human* on purpose: a host runs agent tools from several checkouts, and who answers for that host is a property of the host, not of the directory the tool was launched in.

Two deliberate differences from `deployment-mode.mjs`, both commented in the file:

- **Layer 1 is found by walking UP from cwd**, stopping at the first ancestor with a `.catalyst/config.json`. These identities are read by agent *tools*, which run from a package subdirectory or a `__tests__` folder; a cwd-only check refuses for a repo that is plainly configured one directory down. This is the same walk `secret-contract.mjs`'s `resolveLegacyPerTeamConfigPath` and `linear-comment-post.sh` already do.
- **The id has no fallback; the display name does.** A wrong id mis-files the ask; an absent name is only prose, so the heading reads `Needs from the human`, which is correct on every tenant and reads as deliberate rather than as a placeholder someone forgot.

**2. Wire the two defaults.** `ask.mjs` resolves both the team and the assignee, and does so **before the dry-run returns** — for the same reason the label resolution already sits there: a rehearsal that skips the step which actually fails is not a rehearsal. `--team` stays available (a decision may belong to a team other than the repo's own) but is no longer required. `replica-comment-read.mjs` loses the literal entirely; its default parameter now resolves the tenant's human at call time and throws a named `HumanNotConfiguredError` when there is none — matching that module's existing loud contract, where an unanswerable read is never a clean `null`. `linear-ack.mjs` / `linear-reply.mjs` catch it by name and print a one-line `REFUSED` instead of an unhandled rejection's stack.

**3. The skill text.** Every tenant-shaped literal in `ask`, `concierge` and `steward` references now points at the config key instead: the creating recipe reads both ids out of `.catalyst/config.json` with `jq`, threading names the config key rather than a person, triage says "one human per tenant", the board and asks references say "the workspace owner". Dated provenance attributions (`Ryan, 2026-08-21: "…"`) are kept on purpose — those are citations of who said what, not defaults a customer would inherit, and deleting them destroys the reasoning the rules carry.

**4. `steward/references/cloud-dispatch.md` (new).** On a cloud tenant the dispatch verb is **moving the card to Todo through the write proxy** — `POST /api/v1/agent/issue-state {issueId, stateId, hostId}` — with the two things that route deliberately does not do (it never resolves a state *name*, and a `succeeded` write is not proof work started). Phase-completion evidence is what the tenant wrote down, read from the replica: the `phase.<phase>.complete|failed` outcome comment, the artifact projection, `issues.state` + `issue_history`, `agent_sessions`. No local relay-ticket session is launched — doing so puts two workers on one branch. `dispatch.md` gains a banner pointing here, `SKILL.md`'s dispatch invariant now branches on cloud-detection, and the new file ends with a which-file-governs table.

**5. Config surface.** `catalyst.human.{linearUserId,name}` added to `config.template.json` (null, with the full `$comment` explaining how to obtain it and why there is no default), to this workspace's own `.catalyst/config.json`, and to the configuration reference.

## Tests

`plugins/dev/scripts/lib/tenant-identity.test.mjs` — 16 tests, registered in `execution-core-tests.yml` in this same PR (the four-times-burned unregistered-suite lesson), run under `bun test`, no skips.

Two halves. The ladder half covers every rung including the ones that bite: an empty `ASK_HUMAN_ID` falls through rather than blanking the id, a malformed config is "nothing at this layer" rather than a throw, the upward walk finds Layer 1 from a subdirectory, and **an unconfigured tenant is a named failure whose payload does not contain the old id**.

The second half is the ticket's own acceptance sentence turned into a guard: a **corpus scan** over the shipped skill text and the non-test scripts asserting the fleet owner's Linear user id and the one-person status heading appear nowhere. Per AGENTS.md, a scan that could silently read zero files would report a clean sweep for a repo full of violations — so there is a positive control asserting the walk read a populated corpus (>50 files), that the matcher returns non-zero for a string that *is* present, and that it finds a planted literal in a file of exactly the scanned shape. The scan caught two real violations while I was writing this change, including one in my own new file's header.

Also updated: `replica-comment-read.test.mjs` (the "defaults to the sentinel" test becomes "resolves the configured human / REFUSES when none is configured"), and the two `linear-{ack,reply}-write-path.test.sh` suites now export the identity they seed instead of relying on a default that no longer exists. Their fixture ids are arbitrary uuids now, not a real person's.

## Gate

Run on **mini-2** under the gate lock, against the pushed branch: tenant-identity 16/16, replica-comment-read 16/16, ask-verbs 59/59, linear-ack-write-path 10/10, linear-reply-write-path 38/38, cli-reference-drift 135/135, check-config-drift 51/51, config-template-statemap 6/6, execution-core-config-drift 5/5, skill-shape 161/161.

One step could not run on that host: `plugins/meta/scripts/audit-references.sh` dies at `declare -A` under mini-2's system bash 3.2.57. **Positive control: the identical command fails the same way on mini-2's `main` checkout**, so it is a host limitation, not this branch. It exits 0 on the laptop, and CI runs it on `ubuntu-latest`.

## Scope

Tier 1 only. Tier 2 (the single customer join command) is the cloud side — CTC-493 / CTC-1316 / CTC-1144 — and is not this ticket's half.

### DIFF
--- a/.catalyst/config.json
+++ b/.catalyst/config.json
@@ -1,6 +1,10 @@
 {
   "catalyst": {
     "projectKey": "catalyst-workspace",
+    "human": {
+      "linearUserId": "c2a8cc92-cab6-4536-9500-0f24abdf702b",
+      "name": "Ryan"
+    },
     "project": {
       "ticketPrefix": "CTL"
     },
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -886,6 +886,22 @@ jobs:
         working-directory: plugins/dev/scripts/lib
         run: bun test comment-body-arg.test.mjs
 
+      - name: tenant-identity resolver + fleet-owner-literal scan (bun, CTL-2299)
+        # lib/tenant-identity.mjs resolves WHO the human is, WHICH team an ask is filed on,
+        # and WHAT the status doc calls them — from the tenant's own config, with no
+        # person-shaped fallback. Own step for the same reason as the three leaves above:
+        # it sits outside the execution-core/ tree the "Run stable unit tests" step below is
+        # scoped to, so it needs an explicit run: line or it never executes in CI.
+        #
+        # The suite's second half is a CORPUS SCAN asserting the fleet owner's Linear user
+        # id appears in no shipped skill text and no default — the guard for the rule this
+        # ticket enforces, registered in the SAME change that adds it (the
+        # four-times-burned exec-core-stable-test-allowlist lesson). Positive control:
+        # `grep -n tenant-identity.test .github/workflows/execution-core-tests.yml` returns
+        # this run: line.
+        working-directory: plugins/dev/scripts/lib
+        run: bun test tenant-identity.test.mjs
+
       - name: ask-verb unit tests (bun, CTL-1922 / CTL-1944)
         # ⛔ `ask-verbs.test.mjs` shipped with #3509 and was registered in NO workflow —
         # 20 tests that had never run in CI. Verified with the instrument note above:
--- a/.github/workflows/execution-core-tests.yml
+++ b/.github/workflows/execution-core-tests.yml
@@ -307,10 +307,25 @@ jobs:
       # It was only visible because those suites plant positive controls. A guard that scans for
       # the absence of a token, with no planted control, would have gone GREEN here while checking
       # nothing at all — which is the failure this whole epic is about.
+      # ⛔ `apt-get update` is NOT allowed to be the gate. The hosted runner image ships
+      # third-party apt sources we neither control nor need — and on 2026-09-09 Google's
+      # `dl.google.com/linux/chrome-stable` served a `Packages.gz` whose hash did not match
+      # its own Release file, so `apt-get update -qq` exited 100 and, under `bash -e`, killed
+      # this step before `install ripgrep` ever ran. Every PR in the repo went red on a step
+      # that had not yet done anything, and rerunning did not help because the bad index was
+      # not transient. ripgrep comes from the UBUNTU ARCHIVE, so an unrelated repo's broken
+      # index has no bearing on whether it can be installed.
+      #
+      # The tolerance is scoped precisely: the update is allowed to report errors, the
+      # INSTALL is not, and `rg --version` still has to succeed. So a genuinely unreachable
+      # Ubuntu archive fails this step exactly as loudly as before — which matters, because
+      # the whole reason this step exists is that a missing `rg` makes the scan suites read
+      # an empty tree and pass while checking nothing.
       - name: Install ripgrep (required by the scan-based suites)
         working-directory: .
         run: |
-          sudo apt-get update -qq
+          sudo apt-get update -qq \
+            || echo "::warning::apt-get update reported errors (third-party repo index); continuing — ripgrep comes from the Ubuntu archive and the install below still has to succeed"
           sudo apt-get install -y -qq ripgrep
           rg --version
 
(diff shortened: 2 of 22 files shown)


ryanrozich:
@codex review

ryanrozich:
@codex review

ryanrozich:
[bookkeeping] Automated review could not run — merging on the hand-merge path, recorded rather than claimed.

Codex was triggered twice (on open, and once re-requested after the CI fix) and failed both times with an infrastructure error, not a finding:

    Provided git ref 90a1b5b573d3af15850a226824bbb2e8c699076d does not exist
    Provided git ref 63e158137bf831d9816f611cc7b5da6164f2dbe5 does not exist

Both refs DO exist on origin — `gh api repos/coalesce-labs/catalyst/commits/<sha>` returns each of them — so this is the connector failing to fetch, not a comment on this branch. **A reviewer that could not run is not a clean pass**, so I am not treating it as review evidence and not applying `queue:ready` (which in this repo is the review attestation). Zero review threads exist because none were ever opened.

What the merge is standing on instead: all required checks green at `63e15813` (execution-core-unit-tests, docs-gate, gitleaks, agents-md-gate, audit-references, check-versions, skills-gate, packaging-gate, CodeQL), Mergify Merge Protections satisfied, and a full gate run on mini-2 against this exact sha — tenant-identity 16/16, replica-comment-read 16/16, ask-verbs 59/59, linear-{ack,reply}-write-path 10/10 + 38/38, cli-reference-drift 135/135, check-config-drift 51/51, skill-shape 161/161.

Worth a follow-up: the connector's fetch failure is repo-wide, not specific to this PR.

PR #4140, opened 2026-09-13
fix(dev): CTL-2306 — the thoughts-init forwarder survives a symlink and fails loudly as a lone copy

ryanrozich:
## What

Follow-up to #4139 (CTL-2306 cluster 4d). That PR moved the thoughts layout creator into `plugins/dev/scripts` and left `scripts/worktree-thoughts-init.sh` as a forwarder. The forwarder only worked when it sat in a catalyst checkout and was called by its real path. This PR hardens it.

## Why

Plugin builds before bf76fdd23 find the script at `../../../scripts/` from the plugin's `scripts/`, which in the plugin cache is `~/.claude/plugins/cache/catalyst/scripts/`.
- That file went missing earlier today and was restored by hand. catalyst-cloud's `relay-ticket` skill tells operators to make that copy.
- If someone makes that copy from a checkout at or after bf76fdd23, they copy the forwarder. Its target does not exist there.
- create-worktree hides the init script's output, prints "Could not initialize thoughts", and the CTL-513 guard deletes the new worktree.

## Changes

- **`scripts/worktree-thoughts-init.sh`:**
  - Follows symlinks to its real file (portable loop, bash 3.2 safe) before resolving `../plugins/dev/scripts/worktree-thoughts-init.sh`.
  - When the target is missing, exits 1 with a message naming the canonical `plugins/dev/scripts/worktree-thoughts-init.sh` to copy instead. Previously it fell through to bash's "No such file" and exit 127.
  - Still only forwards; it holds no copy of the layout logic.
- **CI:** `worktree-thoughts-init.test.sh` has never run in CI. It is now registered in `execution-core-tests.yml`, and both path filters include the forwarder and its target.
- **Not in this PR:** the catalyst-cloud `relay-ticket` step moves to the canonical file in a separate catalyst-cloud PR. That is the only fix for plugin builds before bf76fdd23; this forwarder change only makes the wrong copy fail with a readable message.

## Tests (red first)

Commit `684b7f5c2` adds two cases to `worktree-thoughts-init.test.sh`; the suite went from 22 passing / 5 failing before the fix to 27/0 after:
- **Case 7:** runs the forwarder through a relative symlink chain from another directory. It must build the layout.
- **Case 8:** copies the forwarder on its own. It must exit 1, print its own message naming the canonical script (not bash's), and leave no partial layout.

The "thin forwarder" check now asserts that the forwarder has no second copy of the logic, instead of counting lines.

## Local verification

- `worktree-thoughts-init.test.sh` 27/0 under Homebrew bash and `/bin/bash` 3.2.
- `create-worktree-thoughts-check` 26/0, `create-worktree-reconstruct` 11/0.
- `shellcheck` is clean on the forwarder.

**Known local failures, not from this PR:** unchanged from #4139.
- **Same on a clean `origin/main`:** `create-worktree-base-ref` (21/1), `check-project-setup`, `check-project-setup-execution-core`, `catalyst-legacy-plugin`.
- **Same failure sets as `origin/main`:** `compound-log.test.sh`, `briefing-followup-adr-drift`, `briefing-followup-writeback`.
- **Depend on the local environment:** `linear-pr-skip-zsh` Case 3.
- **Stale:** `test-session-instrumentation`.


ryanrozich:
@codex review

PR #4151, opened 2026-09-19
feat(dev): CTC-2819 — fleet shipper defaults move to the stable OTLP name

ryanrozich:
## What

The two fleet-shipper defaults move from the raw tailnet IP `100.65.193.30` to the stable name `otlp.int.catalystcloud.dev`:

- `plugins/dev/scripts/lib/process-memory-metric.mjs:40` — `DEFAULT_COLLECTOR_BASE`
- `plugins/dev/scripts/log-shipper/config.alloy:83` — the `CATALYST_OTLP_ENDPOINT` fallback (and the header comment on line 61 that documents it)

## Why

CTC-2819 moved the OTel collector from `home` to the `clickstack-hel1` box. `otlp.int.catalystcloud.dev` is an A record for whichever host currently runs the collector, so the next host move is a DNS change rather than an edit here.

These two are specifically the defaults a **reprovisioned** fleet host picks up. The running minis have already been moved to the name, but without this change the next provisioning run would put the old address back.

Only the fallback changes — both remain overridable by env (`OTEL_EXPORTER_OTLP_ENDPOINT` / `CATALYST_OTLP_ENDPOINT`), and no test asserts the literal (the existing tests supply their own fixtures).

## Verification

`otlp.int.catalystcloud.dev:4318` accepts OTLP from both minis and the records land in the box's ClickHouse — 5 logs / 3 spans each for markers `ctc2819e-mini-115123` and `ctc2819e-mini-2-115125`, with a live positive control (`claude-code`, 1687 rows in the window) and a negative control (bogus marker, 0 rows). gRPC `:4317` is open from both minis, with `:9999` refused as the negative control.

## Note for the reviewer

`plugins/dev/scripts/execution-core/config.mjs` is deliberately untouched. `COST_CAP_DEFAULT_PROM_URL` points at home's Prometheus, and Prometheus/Loki/Tempo/Grafana all stay on `home` — only the collector and ClickHouse moved.

One consequence worth knowing, not fixed here: `getLokiQueryUrl()` (`config.mjs:1357`) falls back to deriving Loki's address from the OTLP endpoint's host with the port swapped to 3100. That was correct only while the collector and Loki were the same host. Both fleet hosts set `CATALYST_LOKI_QUERY_URL` explicitly in `catalyst-cluster/hosts/*/execution-core.env`, which wins over the fallback, so nothing breaks today — but a **new** host that omits it would now silently resolve Loki to the box, where `:3100` is closed. Worth its own ticket once Loki has a stable name of its own.


ryanrozich:
@codex review

